PR #186, opened 2023-03-06
feat: Add PHP application frame detection

cleptric:
Didn't add tests for Windows paths as Excimer only works on Linux.

As vroom is now a public repo, I also added a changelog 🙂


internal/frame/frame_test.go
```diff
@@ -85,3 +85,49 @@ func TestIsPythonApplicationFrame(t *testing.T) {
...
+				Path: "/var/www/http/vendor/cakephp/cakephp/src/Http/Client.php",
+			},
+			isApplication: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if isApplication := tt.frame.IsPHPApplicationFrame(); isApplication != tt.isApplication {
+				t.Fatalf("Expected %s frame but got %s frame", frameType(tt.isApplication), frameType(isApplication))
```
phacops, line 129: Not sure why you're casting to `frameType` here? Also, I would use `%v` as it will use the default format (`%t` in case of a `bool`).

https://pkg.go.dev/fmt#hdr-Printing

cleptric: Just copied from the test above without putting much thought into it tbh.
Shall I update both tests?

phacops: No, I think we're good actually. I didn't realize we had this `frameType` function to humanize the `bool`. Then `%s` makes sense.

phacops (approved):

PR #194, opened 2023-03-08
feat: Detect an issue when adding source context is slow

phacops:


JonasBa (approved):
Nice, lets get this merged as I'm just interested in seeing how often/what platforms are most impacted but we should be able to remove this soon

PR #199, opened 2023-03-11
feat(storage): provide implementation for other storage providers as profiles storage

### LINKED ISSUE #188: Support for more storage backend providers
Current implementation of vroom is to only support [GCS](https://cloud.google.com/storage/) as the storage backend. It is used multiple times (as `env.profilesBucket`):

* https://github.com/getsentry/vroom/blob/971ab433d1bf57613527add8416179f99a86004c/cmd/vroom/flamegraph.go#L66
* https://github.com/getsentry/vroom/blob/971ab433d1bf57613527add8416179f99a86004c/cmd/vroom/profile.go#L65
* https://github.com/getsentry/vroom/blob/971ab433d1bf57613527add8416179f99a86004c/cmd/vroom/profile.go#L220
* https://github.com/getsentry/vroom/blob/971ab433d1bf57613527add8416179f99a86004c/cmd/vroom/profile.go#L293

It's not suitable for self-hosted version in which will be in GA by late April/beginning of May (per @phacops said). This issue tackles the support for more storage backend providers other than GCS specifically for self-hosted support. 

Some ideas:
- AWS S3 (people should put their bucket configuration on the .env file) - Self-hosted version can use [Minio](https://github.com/minio/minio), but I'd argue that it would take more load to the service stack.
- [BoltDB](https://github.com/boltdb/bolt)/[BadgerDB](https://github.com/dgraph-io/badger) (persisted to disk, can be mounted as normal Docker volume). GCS is basically a KV database, so this would do fine. -- Personally, I support this option more than the first one.

Although I support having BoltDB (or BadgerDB) as the primary storage backend for self-hosted, I think most people who deployed their Sentry instance on cloud services such as AWS, GCP, and Azure should have an option to get a nice and cheap storage bucket. It's nice if we give people more options for that one.

The downside of the above statement is that we have to maintain each implementation. Although during the development process, it's easy because for AWS S3 we have [Minio](https://github.com/minio/minio), Azure we have [Azurite](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio) and we have [some fake-gcs-server Docker image](https://hub.docker.com/r/fsouza/fake-gcs-server) that is used on the [deployment docker-compose file](https://github.com/getsentry/vroom/blob/971ab433d1bf57613527add8416179f99a86004c/deployments/docker-compose.yml#L3-L9). We will still need to assess and maintain the real/live version of storage backend implementations. 

Any feedbacks on this thoughts is appreciated. 

Reference: https://discord.com/channels/621778831602221064/796028405833007104/1081879715645763624

aldy505:
Closes #188 

~~Tests on `storageprovider` package already handled and covered by `storageutils` package. I think there's no need to do the same test on `storageprovider`. But, if you think it'd be better to put additional tests on `storageprovider`, I'll be okay to write it.~~

~~I don't want to support AWS S3 and Azure yet, file storage is fine right now (and far lot cheaper than purchasing separate storage instance for storing profiles data).~~

We're using gocloud.dev/blob now which supports GCS, S3, Azure Blob and Filesystem.


aldy505:
I don't know why the CI is failing at golang-ci lint. Found this thread https://github.com/golangci/golangci-lint/discussions/1920, said I must do `go clean -modcache && go mod tidy`. It doesn't change anything though :/

viglia:
Hey @aldy505 , for a better experience, I recommend installing the `pre-commit` hooks (see the readme) so that you can catch those kind of errors before the CI checks.

It seems what's happening here is that an "import cycle" between `storageutil` and `storageprovider` was introduced: `import cycle not allowed in test`

aldy505:
I've installed pre-commit, but I don't know why it doesn't work on me. Anyway, thanks for the error message. I'll fix the issue later this weekend.

viglia:
> I've installed pre-commit, but I don't know why it doesn't work on me. Anyway, thanks for the error message. I'll fix the issue later this weekend.

@aldy505 After you've installed `pre-commit`, you also need to install the git hook scripts.
I'll fix the doc to add that step as it's missing. Meanwhile you can do that with `pre-commit install`.

With that in place the pre-commit hooks will run each time you commit something.

If you want to manually run the hooks regardless of whether you're committing something or not you can do that with `pre-commit run --all-files`.

phacops:
@aldy505 This is not the direction I want this layer to go towards. I'd rather reuse existing libraries like https://gocloud.dev/howto/blob/ which includes all major cloud providers and a local filesystem option in one package already.

Also, I don't think we'd target support for BadgerDB right now. We need to read and write files on a persistent medium and the filesystem is very well adapted to that for a local option, we don't need an embedded database for that right now.

aldy505:
@phacops I've updated it to use gocloud instead.

internal/storageutil/storageutil.go
```diff
@@ -3,22 +3,31 @@ package storageutil
...
+var ErrObjectNotFound = errors.New("object not found")
+
 // CompressedWrite compresses and writes data to Google Cloud Storage.
-func CompressedWrite(ctx context.Context, b *storage.BucketHandle, objectName string, d interface{}) error {
+func CompressedWrite(ctx context.Context, b *blob.Bucket, objectName string, d interface{}) error {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
-	ow := b.Object(objectName).NewWriter(ctx)
+	ow, err := b.NewWriter(ctx, objectName, &blob.WriterOptions{})
```
phacops, line 23: Was passing `nil` instead of a pointer to an empty `struct` not an option?

internal/storageutil/storageutil_test.go
```diff
@@ -0,0 +1,280 @@
...
+		if err != nil {
+			t.Fatalf("we should be able to marshal back to JSON: %v", err)
+		}
+		if !bytes.Equal(originalData, uncompressedData) {
+			t.Fatalf("data should be identical: %v %v", string(originalData), string(uncompressedData))
+		}
+	})
+}
+
+func TestDownloadProfile_NotFound(t *testing.T) {
```
phacops, line 225: Can you remove the `_` please?

phacops (changes requested):
Thanks again for the contribution. I think it mostly looks good. I'd like you to improve the tests to repeat the code less.

If you look at https://github.com/getsentry/vroom/blob/main/internal/sample/sample_test.go#L14, we define a `struct` containing some data and then apply the same test code for each case. You could follow the same structure to repeat tests on all URLs instead of copy/pasting for each one.

Also, I'm not against launching a `fakestorage` server for the tests only but then you need to remove the one we launch via `docker-compose`.

aldy505:
Also, I'm not against launching afakestorageserver for the tests only but then you need to remove the one we launch viadocker-compose.

@phacops the fakestorage server was there before I opened this PR 🫤

Other than that, I'll do the changes later this weekend.

phacops:
You're right, I set it up that way so we don't have to run `docker-compose` and an additional Docker contain in the tests. I forgot about that.

Then just remove the `docker-compose` config from the repo since we don't need it anymore to run the development environment (we're using the filesystem now).

phacops (approved):

phacops (changes requested):
You need to also get rid of the `docker-compose` invocation in `scripts/run.sh` if you're going to remove the `docker-compose.yml` file.

phacops (approved):

PR #203, opened 2023-03-14
feat: handle is_application in flamegraph payload

eliashussary:
## Summary
Supports `is_application` for flamegraphs.

### DIFF
--- a/internal/flamegraph/flamegraph.go
+++ b/internal/flamegraph/flamegraph.go
@@ -234,6 +234,7 @@ func (f *flamegraph) visitCalltree(node *nodetree.Node, currentStack *[]int) {
 			Name:  node.Name,
 			Image: node.Package,
 			Path:  node.Path,
+			IsApplication: node.IsApplication,
 		}
 		f.framesIndex[frameID] = len(f.frames)
 		*currentStack = append(*currentStack, len(f.frames))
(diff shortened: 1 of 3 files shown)


internal/flamegraph/flamegraph_test.go
```diff
@@ -193,6 +204,19 @@ func TestFlamegraphSpeedscopeGeneration(t *testing.T) {
...
+
+	expectedAppFrames := []speedscope.Frame{
+		{
+			Name:          "c",
+			Image:          "test.package",
+			IsApplication: true,
+		},
+	}
+	actualAppFrames := getApplicationFrames(sp.Shared.Frames)
+	if diff := testutil.Diff(expectedAppFrames, actualAppFrames); diff != "" {
```
phacops, line 216: We should really be calling this once on the whole expected flamegraph `struct`, not each part of the `struct` itself.

That means refactoring above to call `Diff` only once.
```golang
	if diff := testutil.Diff(sp, expectedSp); diff != "" {
```

internal/flamegraph/flamegraph.go
```diff
@@ -234,6 +234,7 @@ func (f *flamegraph) visitCalltree(node *nodetree.Node, currentStack *[]int) {
 			Name:  node.Name,
 			Image: node.Package,
 			Path:  node.Path,
+			IsApplication: node.IsApplication,
```
phacops, line 237: Call https://github.com/getsentry/vroom/blob/333c1d50ed6c4bc9ccde90558651fcf7c5cb84e1/internal/nodetree/nodetree.go#L57 instead to create the frame.

phacops (approved):

PR #209, opened 2023-03-17
ci: add canarychecks to vroom GoCD pipeline

mattgauntseo-sentry:
This script will wait 15 mins to allow for any alerting to trigger before moving to next deployment stage.

In the future we hope to add additional checks, for example - check sentry for alerts / issues.


mattgauntseo-sentry:
@phacops there should be some traffic going to the canary pods and regarding checks we'll likely add to this script, I suspect it will initially be checks using the sentry API (i.e. look for any alerts, spike in issues for the canary pods).

For now, the delay will allow any existing monitoring to trigger pages / alerts that a project might have set up.

mattgauntseo-sentry:
@phacops if you have any concerns over the duration, we can always reduce it to 5 or 10 minutes.

phacops:
I'd be nice to start with 5 minutes yes. I still don't really get what kind of traffic we'll have or even how this will even be able to detect issues with `vroom` in canary so it looks like it'll just be a delay when we want to release until we figure this out.

phacops (approved):

PR #213, opened 2023-03-22
fix: Remove meaningless unsymbolicated before main

phacops:
Fixes https://github.com/getsentry/team-profiling/issues/93

### DIFF
--- a/internal/sample/sample.go
+++ b/internal/sample/sample.go
@@ -593,9 +593,9 @@ func (t *Trace) trimCocoaStacks() {
 		}
 		// Filter unsymbolicated frames after the main frame index
 		ci := msi + 1
-		for fi := msi + 1; fi < len(s)-1; fi++ {
+		for i := ci; i < len(s); i++ {
+			fi := s[i]
 			f := t.Frames[fi]
-			fmt.Println(f)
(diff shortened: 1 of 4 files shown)


internal/sample/sample.go
```diff
@@ -551,3 +564,43 @@ func (p *RawProfile) moveTransaction() {
...
+				break
+			}
+		}
+		// Skip the stack if we're already at the end
+		if msi == len(s)-1 {
+			continue
+		}
+		// Filter unsymbolicated frames after the main frame index
+		ci := msi + 1
+		for fi := msi + 1; fi < len(s)-1; fi++ {
```
Zylphrex, line 596: why `fi < len(s) - 1`? This would leave a frame at the end unchecked right?

internal/sample/sample.go
```diff
@@ -551,3 +564,43 @@ func (p *RawProfile) moveTransaction() {
...
+	}
+	for si, s := range t.Stacks {
+		// Find main frame index in the stack
+		msi := -1
+		for _, fi := range s {
+			if fi == mfi {
+				msi = fi
+				break
+			}
+		}
```
Zylphrex, line 589: Nit: Would this be slightly more performant if you iterated the stack backwards because `main` is always close to the end?

PR #219, opened 2023-03-23
fix: Compare the frame index instead of the index in the stack

phacops:


internal/sample/sample.go
```diff
@@ -584,15 +584,21 @@ func (t *Trace) trimCocoaStacks() {
 	}
 	for si, s := range t.Stacks {
 		// Find main frame index in the stack
-		msi := -1
-		for fi := len(s) - 1; fi >= 0; fi-- {
+		msi := len(s)
+		// Stop searching after 10 frames, it's not there
+		var until int
```
Zylphrex, line 589: I assume this gets a default value of 0?

Zylphrex (approved):

internal/sample/sample.go
phacops: Yes.

PR #224, opened 2023-03-24
fix(profiling): Do not update package on any platform

Zylphrex:
Package should only be used for native platforms like cocoa and rust. This only updates it on the frame and not the nodetree because the nodetree is serialized for the suspect functions feature which coalesces module and package together still so that is left untouched. This also ensures that the frame fingerprint does not change.

### DIFF
--- a/internal/frame/frame.go
+++ b/internal/frame/frame.go
@@ -83,7 +83,9 @@ func (f Frame) PackageBaseName() string {
 
 func (f Frame) WriteToHash(h hash.Hash) {
 	var s string
-	if f.Package != "" {
+	if f.Module != "" {
+		s = f.Module
+	} else if f.Package != "" {
 		s = f.PackageBaseName()
 	} else if f.File != "" {
 		s = f.File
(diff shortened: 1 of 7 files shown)


internal/frame/frame.go
```diff
@@ -72,19 +72,19 @@ func (f Frame) ID() string {
...
 }
 
 func (f Frame) WriteToHash(h hash.Hash) {
 	var s string
-	if f.Package != "" {
-		s = f.PackageBaseName()
+	if f.Module != "" {
+		s = f.Module
+	} else if f.Package != "" {
+		s = path.Base(f.Package)
```
Zylphrex, line 87: This will preserve the function fingerprints, do we care about that?

internal/nodetree/nodetree.go
```diff
@@ -56,7 +56,6 @@ func (n *Node) Update(timestamp uint64) {
 
 func (n *Node) ToFrame() frame.Frame {
 	n.Frame.Data.SymbolicatorStatus = n.Frame.Status
-	n.Frame.Package = n.Frame.PackageBaseName()
```
Zylphrex, line 59: This preserves the original package name

phacops (approved):

internal/frame/frame.go
phacops: I don't think we do but it definitely won't hurt. Less disruption for a small change.

PR #237, opened 2023-03-31
feat: Set package and in app status for node frames

phacops:


Zylphrex (approved):

internal/occurrence/occurrence.go
```diff
@@ -142,7 +142,7 @@ func NewOccurrence(p profile.Profile, ni nodeInfo) *Occurrence {
 		"frame_duration_ns":   ni.Node.DurationNS,
 		"frame_module":        ni.Node.Frame.Module,
 		"frame_name":          ni.Node.Name,
-		"frame_package":       ni.Node.Frame.Package,
+		"frame_package":       ni.Node.Frame.ModuleOrPackage(),
```
Zylphrex, line 145: Why is this changing?

internal/sample/sample.go
```diff
@@ -615,3 +616,17 @@ func (t *Trace) trimCocoaStacks() {
...
 	}
 }
+
+func (t *Trace) normalizeNodeFrames() {
+	packageInfo := make(map[string]packageutil.PackageInfo)
+	for i := 0; i < len(t.Frames); i++ {
+		f := t.Frames[i]
+		pi, exists := packageInfo[f.Path]
+		if !exists {
+			pi = packageutil.ParseNodeModuleFromPath(f.Path)
```
Zylphrex, line 626: did you mean to add this to the `packageInfo` map but forgot?

internal/occurrence/occurrence.go
phacops: It shouldn't indeed.

internal/sample/sample.go
phacops: Yes.

PR #249, opened 2023-04-26
fix: Do not send occurrences without tag as they are considered invalid by the issue platform

phacops:


internal/occurrence/kafka.go
```diff
@@ -9,6 +9,11 @@ import (
 func GenerateKafkaMessageBatch(occurrences []*Occurrence) ([]kafka.Message, error) {
 	messages := make([]kafka.Message, 0, len(occurrences))
 	for _, o := range occurrences {
+		// Skip event with no tag as they are considered invalid by the issue platform
+		if len(o.Event.Tags) == 0 {
+			continue
+		}
```
Zylphrex, line 15: I'm not 100% sure but is it possible for a transaction with no tags to be sent? That would lead to this case as well. Should we just send the occurrence with an empty map for the tags instead?

Zylphrex (approved):

PR #261, opened 2023-05-19
ref: receive a list of profiles ID for flamegraph generation

viglia:
Instead of querying snuba for a list of profiles ID, we'll be receiving a list of `profiles_id` directly from `sentry`.

I'm turning this into a `POST` endpoint as we'll be getting a list of profiles id plus (in the near future) an optional list (for each profile id) of span data (start, end) so URL parameters might not be ideal.

Once the `sentry` change will be deployed too, I'll open another PR to remove the unused code/endpoint.

### DIFF
--- a/cmd/vroom/flamegraph.go
+++ b/cmd/vroom/flamegraph.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"io"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/vroom/flamegraph.go
```diff
@@ -85,3 +86,75 @@ func (env *environment) getFlamegraph(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 	_, _ = w.Write(b)
 }
+
+type getFlamegraphFromProfileIDs struct {
+	ProfilesID []string `json:"profiles_id"`
+}
+
+func (env *environment) getFlamegraphFromProfileIDs(w http.ResponseWriter, r *http.Request) {
```
phacops, line 94: Let's call that `postFlamegraphFromProfileIDs`.

cmd/vroom/flamegraph.go
```diff
@@ -85,3 +86,75 @@ func (env *environment) getFlamegraph(w http.ResponseWriter, r *http.Request) {
...
+	var profiles getFlamegraphFromProfileIDs
+	s := sentry.StartSpan(ctx, "json.unmarshal")
+	s.Description = "Unmarshal list of profiles"
+	err = json.Unmarshal(body, &profiles)
+	s.Finish()
+	if err != nil {
+		hub.CaptureException(err)
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
```
phacops, line 134: Let's use a `json.NewDecoder` and stream it through.

cmd/vroom/flamegraph.go
```diff
@@ -85,3 +86,75 @@ func (env *environment) getFlamegraph(w http.ResponseWriter, r *http.Request) {
...
+
+	hub.Scope().SetTag("fetched_profiles", strconv.Itoa(len(profiles.ProfilesID)))
```
phacops, line 136: That would not be the right thing reported. This is the number of profiles we sent, not the actual number we fetched. This should be lower down or removed.

phacops (changes requested):

cmd/vroom/flamegraph.go
```diff
@@ -85,3 +86,75 @@ func (env *environment) getFlamegraph(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 	_, _ = w.Write(b)
 }
+
+type getFlamegraphFromProfileIDs struct {
+	ProfilesID []string `json:"profiles_id"`
```
phacops, line 91: `ProfileIDs`

phacops (approved):

PR #264, opened 2023-05-30
feat: refactor flamegraph endpoint  to allow for an optional list of span iintervals

viglia:
This PR refactors the `flamegraph` endpoint to allow for an optional list of span intervals for each profile.

If span intervals are present, the relative profile will be sliced accordingly.

### DIFF
--- a/internal/flamegraph/span_util.go
+++ b/internal/flamegraph/span_util.go
@@ -87,5 +87,8 @@ func overlappingDuration(node *nodetree.Node, interval *SpanInterval) uint64 {
 	end := min(node.EndNS, interval.End)
 	start := max(node.StartNS, interval.Start)
 
-	return max(0, (end - start))
+	if end <= start {
+		return 0
+	}
+	return end - start
 }


internal/flamegraph/span_util.go
```diff
@@ -0,0 +1,94 @@
...
+			sampleCount := math.Ceil(float64(duration) / float64(time.Millisecond*10))
+			node.SampleCount = int(sampleCount)
+			if children := sliceCallTree(&node.Children, intervals); len(children) > 0 {
+				node.Children = children
+			} else {
+				node.Children = nil
+			}
+			slicedTree = append(slicedTree, node)
+		} else {
+			continue
```
Zylphrex, line 63: This else block is unnecessary

internal/flamegraph/span_util.go
```diff
@@ -0,0 +1,94 @@
...
+
+func mergeIntervals(intervals *[]SpanInterval) []SpanInterval {
+	sort.SliceStable((*intervals), func(i, j int) bool {
+		if (*intervals)[i].Start == (*intervals)[j].Start {
+			return (*intervals)[i].End < (*intervals)[j].End
+		}
+		return (*intervals)[i].Start < (*intervals)[j].Start
+	})
+
+	newIntervals := []SpanInterval{(*intervals)[0]}
```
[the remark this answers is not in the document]
viglia: yes, we only fetch profiles associated to transactions that have a specific span group (in case it's a span scoped request) so there should always be at least one interval.

With that said, I agree that I should add an extra security-check. After all, a span ingestion in snuba might fail for whatever reason and in that case we'd find ourselves with a transaction claiming to have a given span but then no span in the `spans` table.

cmd/vroom/flamegraph.go
```diff
@@ -87,7 +87,8 @@ func (env *environment) getFlamegraph(w http.ResponseWriter, r *http.Request) {
 }
 
 type postFlamegraphFromProfileIDs struct {
-	ProfileIDs []string `json:"profile_ids"`
+	ProfileIDs []string                     `json:"profile_ids"`
+	Spans      *[][]flamegraph.SpanInterval `json:"spans,omitempty"`
```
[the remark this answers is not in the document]
viglia: correct. I've just added a few line of comments to make this clear.

internal/flamegraph/span_util.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return b
+}
+
+func sliceCallTree(callTree *[]*nodetree.Node, intervals *[]SpanInterval) []*nodetree.Node {
+	slicedTree := make([]*nodetree.Node, 0)
+	for _, node := range *callTree {
+		if duration := getTotalOverlappingDuration(node, intervals); duration > 0 {
+			sampleCount := math.Ceil(float64(duration) / float64(time.Millisecond*10))
+			node.SampleCount = int(sampleCount)
```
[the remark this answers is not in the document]
viglia: > I guess there's no way to get the actual sample count right now because we lose the sample timestamps in the call tree right?

yes, correct. If we were to deal just with the sample format then we could just have a method to slice the profile by dropping any sample that does fall within any of the span intervals before creating the call tree but this is not the case.


> I raise this because we've noticed repeatedly that our sampling interval isn't consistently 10ms (we actually aim for a sampling frequency of 101hz = 1000ms/101 = 9.9009ms). But in reality, the sampling mechanism can take much longer before taking a sample.

yes, I'm aware of this. Not having specific sample timestamps, we're estimating it based on the assumption of samples being collected every 10 ms.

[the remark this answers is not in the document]
viglia: @Zylphrex as per Slack discussion, I'm marking this as resolved while keeping in mind that if we ever need a better precision then we can support counting the right sample count by span at CallTree construction time.

cmd/vroom/flamegraph.go
```diff
@@ -135,15 +132,7 @@ func (env *environment) postFlamegraphFromProfileIDs(w http.ResponseWriter, r *h
...
-	speedscope, err := flamegraph.GetFlamegraphFromProfiles(
-		ctx,
-		env.storage,
-		organizationID,
-		projectID,
-		profiles.ProfileIDs,
-		numWorkers,
-		timeout,
-	)
+	speedscope, err := flamegraph.GetFlamegraphFromProfiles(ctx, env.storage, organizationID, projectID, profiles.ProfileIDs, profiles.Spans, numWorkers, timeout)
```
Zylphrex, line 135: Would be good to have an assertion that `profiles.ProfileIDS` and `profiles.Spans` have the same length and return a 400.

internal/flamegraph/span_util.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, interval := range *intervals {
+		if node.EndNS <= interval.Start {
+			// in this case any remaining interval
+			// starts after the given call frame
+			// therefeore we can bail out early
+			break
+		}
+		if overlap(node, &interval) {
+			duration += overlappingDuration(node, &interval)
+		}
```
Zylphrex, line 85: This is more of a nit but `overlap` and `overlappingDuration` seems to be almost identical. One calculates if `a > b` and the other calculates `a - b`. I wonder if you can change `overlappingDuration` to just return 0 when `a <= b`.

Zylphrex (approved):

cmd/vroom/flamegraph.go
viglia: This is taken care in `sentry` where I'll ensure that if `profiles.Spans` is passed then it'll be the same length (having empty lists if necessary) but yes, I can add a further check here regardless of what happens in sentry.

internal/flamegraph/span_util.go
```diff
@@ -91,5 +87,5 @@ func overlappingDuration(node *nodetree.Node, interval *SpanInterval) uint64 {
 	end := min(node.EndNS, interval.End)
 	start := max(node.StartNS, interval.Start)
 
-	return (end - start)
+	return max(0, (end - start))
```
Zylphrex, line 90: I dont think you can do this, because if `start < end` this will underflow since it's an `uint64`

PR #267, opened 2023-06-02
feat(android): Use app identifier to improve android frame classifica…

Zylphrex:
…tion

All application frames should have the app identifier as the prefix on the package information. This change uses the app identifier information when it's available to improve the classification of frames as application vs system.

### DIFF
--- a/internal/profile/legacy.go
+++ b/internal/profile/legacy.go
@@ -128,9 +128,7 @@ func (p *LegacyProfile) UnmarshalJSON(b []byte) error {
 		// on the android profile as it will be used when classifying
 		// frames as application vs system
 		metadata := p.GetTransactionMetadata()
-		if metadata.AppIdentifier != "" {
-			t.AppIdentifier = &metadata.AppIdentifier
-		}
+		t.AppIdentifier = metadata.AppIdentifier
 
 		p.Trace = &t
 		p.Profile = nil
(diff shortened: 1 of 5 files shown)


internal/profile/legacy.go
```diff
@@ -123,6 +123,15 @@ func (p *LegacyProfile) UnmarshalJSON(b []byte) error {
...
 			return err
 		}
+
+		// if the app identifier is present, then make sure to set it
+		// on the android profile as it will be used when classifying
+		// frames as application vs system
+		metadata := p.GetTransactionMetadata()
+		if metadata.AppIdentifier != "" {
+			t.AppIdentifier = &metadata.AppIdentifier
+		}
```
Zylphrex, line 133: This seems like a pretty evil hack. But the alternative is to pass something to the `CallTrees` or `Speedscope` functions. Open to other ideas.

internal/profile/legacy.go
```diff
@@ -123,6 +123,13 @@ func (p *LegacyProfile) UnmarshalJSON(b []byte) error {
 		if err != nil {
 			return err
 		}
+
+		// if the app identifier is present, then make sure to set it
+		// on the android profile as it will be used when classifying
+		// frames as application vs system
+		metadata := p.GetTransactionMetadata()
+		t.AppIdentifier = metadata.AppIdentifier
```
Zylphrex, line 131: This seems like a pretty evil hack. But the alternative is to pass something to the `CallTrees` or `Speedscope` functions so we have access to the app identifier. Open to other ideas.

Zylphrex:
closing in favour of #273

PR #269, opened 2023-06-06
mitigate sample count estimation error

viglia:
In some cases the sampling frequency is not respected.

This is particularly true for php, where the gap between two samples could be up to 30s and for Python, in case there is some native code that is holding the GIL.

This aim at mitigating (not fixing) the potential error estimating sample counts in such scenario.

The idea is that the sample count estimated for a node given a certain interval that overlaps it, should never be more than the original sample count.

So, for example, given a node with a total duration of `30s` but a sample count of `2`, if there is an interval that overlaps for the whole `30s`, instead of estimating a sample count of `3,000`, we'll set it to `2`.

### DIFF
--- a/internal/flamegraph/span_util.go
+++ b/internal/flamegraph/span_util.go
@@ -55,6 +55,10 @@ func sliceCallTree(callTree *[]*nodetree.Node, intervals *[]SpanInterval) []*nod
 	for _, node := range *callTree {
 		if duration := getTotalOverlappingDuration(node, intervals); duration > 0 {
 			sampleCount := math.Ceil(float64(duration) / float64(time.Millisecond*10))
+			// here we take the minimum between the node sample count and the estimated
+			// sample count to mitigate the case when we make a wrong estimation due
+			// to sampling frequency not being respected. (Python native code holding
+			// the GIL, php, etc.)
 			node.SampleCount = int(min(uint64(sampleCount), uint64(node.SampleCount)))
 			if children := sliceCallTree(&node.Children, intervals); len(children) > 0 {
 				node.Children = children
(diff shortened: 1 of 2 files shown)


Zylphrex (approved):

internal/flamegraph/span_util.go
```diff
@@ -55,7 +55,11 @@ func sliceCallTree(callTree *[]*nodetree.Node, intervals *[]SpanInterval) []*nod
 	for _, node := range *callTree {
 		if duration := getTotalOverlappingDuration(node, intervals); duration > 0 {
 			sampleCount := math.Ceil(float64(duration) / float64(time.Millisecond*10))
-			node.SampleCount = int(sampleCount)
+			// here we take the minimum between the node sample count and the estimated
+			// sample count to mitigate the case when we make a wrong estimation due
+			// to sampling frequency not being respected. (Python native code holding
+			// the GIL, php, etc.)
+			node.SampleCount = int(min(uint64(sampleCount), uint64(node.SampleCount)))
```
phacops, line 62: We should pass a function taking a node and a float (the `sampleCount` calculated above) and this function will return the right sample count.

Now, this function would be created around here https://github.com/getsentry/vroom/blob/c160bd8b2b2e0947e1e054216fbca0be4550da45/internal/flamegraph/flamegraph.go#L83 and based on the platform, take the exact count (for sample format profiles) or an estimation (for android).

The only reason we created this estimation was because we didn't have sample count for android. We do not want to "estimate" at all when we have an exact sample count.

phacops (changes requested):

phacops (approved):
@Zylphrex reminded me of the details, it's all good now.

PR #272, opened 2023-06-06
ref(legacy): Make legacy profile the promoted type

Zylphrex:
By making the legacy profile the promoted type, this allows us the make the legacy android + ios profile types the top level type, giving us access to the various metadata fields when we work with the raw data.

This also has the added benefit of decoupling the legacy ios and android profiles so the legacy ios profile can be easily removed when we drop support for it.

### DIFF
--- a/internal/profile/android.go
+++ b/internal/profile/android.go
@@ -1,6 +1,7 @@
 package profile
 
 import (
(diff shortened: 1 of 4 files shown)


Zylphrex:
closing in favour of #273

PR #273, opened 2023-06-07
feat(android): Normalize android methods to contain in app

Zylphrex:
Introduces a new in app field on the android method that will be used to store the precomputed result of application vs system frame.

### DIFF
--- a/internal/packageutil/package.go
+++ b/internal/packageutil/package.go
@@ -45,11 +45,7 @@ var (
 )
 
 // IsAndroidApplicationPackage checks if a symbol belongs to an Android system package.
-func IsAndroidApplicationPackage(packageName string, appIdentifier string) bool {
-	if appIdentifier != "" {
-		return strings.HasPrefix(packageName, appIdentifier+".")
-	}
-
+func IsAndroidApplicationPackage(packageName string) bool {
 	for _, p := range androidPackagePrefixes {
 		if strings.HasPrefix(packageName, p) {
 			return false
--- a/internal/packageutil/package.go
+++ b/internal/packageutil/package.go
@@ -45,7 +45,11 @@ var (
 )
 
 // IsAndroidApplicationPackage checks if a symbol belongs to an Android system package.
-func IsAndroidApplicationPackage(packageName string) bool {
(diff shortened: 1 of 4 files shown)


internal/packageutil/package.go
```diff
@@ -45,7 +45,11 @@ var (
 )
 
 // IsAndroidApplicationPackage checks if a symbol belongs to an Android system package.
-func IsAndroidApplicationPackage(packageName string) bool {
+func IsAndroidApplicationPackage(packageName string, appIdentifier string) bool {
+	if appIdentifier != "" {
```
phacops, line 49: I don't think we need this anymore.

phacops (changes requested):

internal/profile/android.go
```diff
@@ -251,6 +257,20 @@ func generateFingerprint(stack []*nodetree.Node) uint64 {
...
 }
 
+func (p *Android) NormalizeMethods(pi profileInterface) {
+	metadata := pi.GetTransactionMetadata()
+	appIdentifier := metadata.AppIdentifier
+
+	for i := range p.Methods {
+		method := p.Methods[i]
+
+		inApp := packageutil.IsAndroidApplicationPackage(method.ClassName, appIdentifier)
```
phacops, line 267: I would just do the `HasPrefix` here and forget about this method.

internal/profile/android.go
```diff
@@ -279,12 +310,18 @@ func (p Android) Speedscope() (speedscope.Output, error) {
 			if err != nil {
 				return speedscope.Output{}, err
 			}
+			var inApp bool
+			if method.InApp != nil {
+				inApp = *method.InApp
+			} else {
+				inApp = packageutil.IsAndroidApplicationPackage(method.ClassName)
```
Zylphrex, line 317: This particular call previously used `packageName` which strips the segment after the final `.`. Not sure if this actually makes a difference but if a function was like `com.example.foo.Bar()` then it would actually not get detected correctly. Just updating it here to use the `.ClassName` to be consistent with all the other instances.

phacops (approved):

PR #274, opened 2023-06-07
feat(nodejs): Classify node internals as system frames

Zylphrex:
Node internal frames have abs_path that start with `node:internal`. Correctly classify these as system frames.


JonasBa (approved):
🙏🏼

PR #280, opened 2023-06-15
fix: replace unix epoch timestamp of span intervals with ns elapsed since the start of the profile

viglia:
This is to be consistent with what is already being used in both the sample and the node structs.

### DIFF
--- a/internal/flamegraph/span_util.go
+++ b/internal/flamegraph/span_util.go
@@ -9,8 +9,8 @@ import (
 )
 
 type SpanInterval struct {
-	Start uint64 `json:"start"`
-	End   uint64 `json:"end"`
+	Start uint64 `json:"start,string"`
+	End   uint64 `json:"end,string"`
 }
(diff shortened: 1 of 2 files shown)


internal/flamegraph/span_util.go
```diff
@@ -36,6 +36,24 @@ func mergeIntervals(intervals *[]SpanInterval) []SpanInterval {
 	return newIntervals
 }
 
+func relativeIntervalsFromAbsoluteTimestamp(intervals *[]SpanInterval, t uint64) {
+	for i := range *intervals {
```
phacops, line 40: I think it'd be nicer to get both the index and the value here and use the value where appropriate and only the index when it's required to modify the value.

Also, you could use a 2nd loop to go over `Start` and `End` for each and use the `max` function defined below to not have to do 2 conditions.

phacops (approved):

PR #284, opened 2023-06-25
ref: cleanup binary for self-hosted

### LINKED ISSUE #268: Create a cleanup job based on specified retention days
Regarding https://github.com/getsentry/self-hosted/pull/2154#discussion_r1210881838:

> BYK It's definitely an issue we didn't really think about since we were relying on GCS retention settings to remove profiles when needed. I should be addressed separately in vroom.

On sentry and snuba image, they both respect `SENTRY_EVENT_RETENTION_DAYS` environment variable in which will automatically delete the data longer than the provided retention days. While the self-hosted version of vroom rely on filesystem, should we create a simple goroutine (instead of having it as a separate process -- which will add a new container on the self-hosted docker-compose) that will do a cleanup job everyday?

aldy505:
This is a PR that tackles https://github.com/getsentry/self-hosted/pull/2211#issuecomment-1596993724 and https://github.com/getsentry/self-hosted/pull/2211#issuecomment-1601622430 on event that people at Sentry don't want to use an external cron scheduler, and wanted to put it as a vroom extension (of sorts).

I don't think I'll have any free time during next two weeks, meaning I'll be late to response. To mitigate that, I'll just create this one and let you guys decide it.

To apply this on `self-hosted` repo, I (or someone) just need to modify this on the `docker-compose.yml` file:
```yaml
  vroom-cleanup:
    <<: *restart_policy
    image: "$VROOM_IMAGE"
    entrypoint: "/bin/cleanup"
    environment:
      SENTRY_BUCKET_PROFILES: file://localhost//var/lib/sentry-profiles
      # Leaving the value empty to just pass whatever is set
      # on the host system (or in the .env file)
      SENTRY_EVENT_RETENTION_DAYS:
    volumes:
      - sentry-vroom:/var/lib/sentry-profiles
```

Closes #268 

### Legal Boilerplate

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,8 @@ COPY . .
 
 RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o . -ldflags="-s -w -X main.release=$(git rev-parse HEAD)" ./cmd/vroom
 
+RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o . -ldflags="-s -w" ./cmd/cleanup
+
 FROM alpine
 
 EXPOSE 8080
@@ -17,6 +19,8 @@ RUN apk add --no-cache ca-certificates tzdata && mkdir -p $PROFILES_DIR
 ENV SENTRY_BUCKET_PROFILES=file://localhost/$PROFILES_DIR
 
 COPY --from=builder /src/vroom /bin/vroom
+
+COPY --from=builder /src/cleanup /bin/cleanup
 
 WORKDIR /var/vroom
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 **
 !/.git
 !/cmd/vroom
+!/cmd/cleanup
 !/internal
 !/pkg
 !/vendor
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /vroom
 /downloader
 /issuedetection
+/cleanup
 
 # Test binary, built with `go test -c`
 *.test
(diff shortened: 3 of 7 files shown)


Dockerfile
```diff
@@ -6,6 +6,8 @@ COPY . .
 
 RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o . -ldflags="-s -w -X main.release=$(git rev-parse HEAD)" ./cmd/vroom
 
+RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o . -ldflags="-s -w" ./cmd/cleanup
```
phacops, line 9: We don't need this on another line, we can add `./cmd/cleanup` to the line above.

Dockerfile
```diff
@@ -18,6 +20,8 @@ ENV SENTRY_BUCKET_PROFILES=file://localhost/$PROFILES_DIR
 
 COPY --from=builder /src/vroom /bin/vroom
 
+COPY --from=builder /src/cleanup /bin/cleanup
```
phacops, line 23: We should change the `COPY` above to include `/src/cleanup` instead.

phacops (commented):
I'm not sure why we'd need a dedicated binary to do that when this command will do the same thing and be implemented nicely in `self-hosted`:
```sh
find ${SENTRY_BUCKET_PROFILES} -type f -mtime +${SENTRY_EVENT_RETENTION_DAYS} -delete
```

For now, I don't think we should merge this.

aldy505:
@phacops Please see https://github.com/getsentry/self-hosted/pull/2211#issuecomment-1601622430

aldy505:
@phacops regarding https://github.com/getsentry/self-hosted/pull/2211#issuecomment-1636671167, can you please close this PR when you've migrated vroom to Debian? Or should I also do the base image migration?

aldy505:
Closing in favor of #290

PR #290, opened 2023-07-17
Use debian bullseye in DockerFile

hubertdeng123:
I'll test this once image is built to see if it works without issues for self-hosted

#skip-changelog

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@ EXPOSE 8080
 
 ARG PROFILES_DIR=/var/lib/sentry-profiles
 
-RUN apk add --no-cache ca-certificates tzdata && mkdir -p $PROFILES_DIR
+RUN apt-get install ca-certificates tzdata && mkdir -p $PROFILES_DIR && \
+    rm -r /var/lib/apt/lists/*
 
 ENV SENTRY_BUCKET_PROFILES=file://localhost/$PROFILES_DIR
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,11 @@ EXPOSE 8080
 
 ARG PROFILES_DIR=/var/lib/sentry-profiles
 
-RUN apt-get update && apt-get install -y ca-certificates tzdata && \
+RUN apt-get update \
+    && apt-get install -y ca-certificates tzdata --no-install-recommends \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && mkdir -p $PROFILES_DIR
     mkdir -p $PROFILES_DIR && \
     rm -r /var/lib/apt/lists/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@ EXPOSE 8080
 
 ARG PROFILES_DIR=/var/lib/sentry-profiles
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,13 +6,15 @@ COPY . .
 
 RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o . -ldflags="-s -w -X main.release=$(git rev-parse HEAD)" ./cmd/vroom
 
-FROM alpine
+FROM debian:bullseye-slim
```
aldy505, line 9: Bookworm is the latest stable now. Shouldn't we use that instead?

hubertdeng123: the reason why I'm using bullseye is because symbolicator and sentry are both using that, so that's what I would prefer

Dockerfile
```diff
@@ -6,13 +6,15 @@ COPY . .
...
 
 EXPOSE 8080
 
 ARG PROFILES_DIR=/var/lib/sentry-profiles
 
-RUN apk add --no-cache ca-certificates tzdata && mkdir -p $PROFILES_DIR
+RUN apt-get update && apt-get install -y ca-certificates tzdata && \
```
phacops, line 15: I suggest we add an `apt-get clean` and `--no-install-recommends` as well:
```suggestion
RUN apt-get update \
    && apt-get install -y ca-certificates tzdata --no-install-recommends \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/* \
    && mkdir -p $PROFILES_DIR
```

Dockerfile
phacops: I think I'd rather have you use the latest stable. I don't think we have an advantage keeping things the same between `symbolicator` and `vroom`.

phacops (changes requested):

Dockerfile
aldy505: I'd also rather use bookworm as it has better security packages rather than bullseye (in which my highest concern is actually OpenSSL v1.1 on bullseye which [will be EOL within 1 month](https://endoflife.date/openssl)).

But considering these two things, I'd use bullseye instead:
- Other images are currently using bullseye. If we're about to use to bookworm, I'd prefer other images to migrate too.
- Do the SRE and infrastructure team at Sentry able to maintain bookworm? If they don't have the confidence to move to newer Debian release, I'd rather wait until they're confident enough.

hubertdeng123: >  I don't think we have an advantage keeping things the same between symbolicator and vroom.

In the past, we've seen people experience some frustration around having different versions of debian for our containers. Having the same version would ease personalization as described in the issue here.
https://github.com/getsentry/self-hosted/issues/1128

phacops: I understand the frustration of the users but as it's pointed out in the issue, maybe the solution to bundle `cron` is not such a bad idea.

I don't like cross-project dependencies between projects because of another project's cleanup strategy.

phacops (approved):

Dockerfile
hubertdeng123: > I understand the frustration of the users but as it's pointed out in the issue, maybe the solution to bundle cron is not such a bad idea.

Yeah, you may be right. Can revisit this later

PR #292, opened 2023-07-18
Add Android Regex symbols

stefanosiano:
Added new symbols for regex on main thread on Android


armcknight (approved):
LGTM, just needs the changelog as mentioned

PR #297, opened 2023-08-02
feat: Send occurrences on frozen frame drop

phacops:
This PR will find the stack likely responsible for the frozen frame drop and, if the frame drop originates from an in app function, will report an occurrence.

### DIFF
--- a/internal/occurrence/frame_drop.go
+++ b/internal/occurrence/frame_drop.go
@@ -32,7 +32,6 @@ func findFrameDropCause(
 			)
 			// We found a potential stacktrace responsible for this frozen frame
 			if frameDropInfo != nil {
-				// Create occurrences.
 				stackTrace := make([]frame.Frame, 0, len(frameDropInfo.StackTrace))
 				for _, f := range frameDropInfo.StackTrace {
 					stackTrace = append(stackTrace, f.ToFrame())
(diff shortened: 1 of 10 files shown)


internal/occurrence/frame_drop.go
```diff
@@ -0,0 +1,135 @@
...
+		// If the function we'd like to return is unknown, the issue is not
+		// actionable and we drop it.
+		if node.Frame.Function == "" {
+			return nil
+		}
+		fdi := frameDropInfo{
+			Node:       *node,
+			StackTrace: make([]*nodetree.Node, len(stackTrace)),
+		}
+		fdi.Node.Children = nil
```
Zylphrex, line 98: Won't this modify the original call tree? And since issue detection happens before function aggregation, function aggregation might break.

internal/occurrence/frame_drop.go
```diff
@@ -0,0 +1,135 @@
...
+		return candidates[i].durationNS > candidates[j].durationNS
+	})
+	// We pick the child with the longest duration and continue to look for
+	// the leaf.
+	return findFrameDropFrame(
+		n.Children[candidates[0].index],
+		frozenFrameStartNS,
+		frozenFrameEndNS,
+		st,
+	)
```
Zylphrex, line 134: This is more of a product question but suppose we had a call tree like this

```
[ a                                     ]
[ b                          ][ c       ]
[ d ][ e ][ d ][ e ][ d ][ e ][ f       ]
```

And the frozen frame lasted for the duration of `a` what would you want to show as the cause of the frame drop?

internal/occurrence/frame_drop.go
Zylphrex: Already being copied, so all good

Zylphrex (approved):

internal/occurrence/frame_drop.go
phacops: Yes, that's a good point. I don't have enough data right now to be able to see if a more complicated algorithm would be better. This way of selecting the longest function at each level is a way to get started on reporting the data we see.

PR #300, opened 2023-08-03
feat: use profiles that were not dynamically sampled to enhance slowest functions aggregation

viglia:
`getsentry` [PR](https://github.com/getsentry/getsentry/pull/11335)

`sentry` [PR](https://github.com/getsentry/sentry/pull/54023)

`relay` [PR](https://github.com/getsentry/relay/pull/2940)

This will let us use profiles that are not dynamically sampled to enhance the `slowest functions` aggregation with more data.

⚠ this should only be merged after the `sentry` PR is merged.


cmd/vroom/profile.go
```diff
@@ -71,24 +72,33 @@ func (env *environment) postProfile(w http.ResponseWriter, r *http.Request) {
...
+				// These errors won't be retried
+				hub.CaptureException(err)
+				if code := gcerrors.Code(err); code == gcerrors.FailedPrecondition {
+					w.WriteHeader(http.StatusPreconditionFailed)
+				} else {
+					w.WriteHeader(http.StatusInternalServerError)
+				}
 			}
 		}
-		return
```
Zylphrex, line 91: Why was this early return removed?

cmd/vroom/profile.go
```diff
@@ -102,35 +112,37 @@ func (env *environment) postProfile(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if len(callTrees) > 0 {
-		s = sentry.StartSpan(ctx, "processing")
-		s.Description = "Find occurrences"
-		occurrences := occurrence.Find(p, callTrees)
-		s.Finish()
+		if p.IsSampled() {
```
Zylphrex, line 115: A comment explaining why we only look for occurrences on sampled profiles would be helpful here

cmd/vroom/profile.go
viglia: not on purpose. Good catch!

Zylphrex (approved):

PR #306, opened 2023-09-05
feat: Accept and return the time a profile started at in a timestamp field on Android

phacops:


Zylphrex (approved):
I think you also have to add this field to relay right?

PR #308, opened 2023-09-08
feat: Filter frame drop candidates based on thresholds

phacops:
Frame draps are supposed to happened at the beginning of the frozen frame and covering a good portion of the duration of the frozen frame. In this PR, I added 2 filters to weed out bad candidates and loosened when the start and end of the frame drop cause can be.

Now we'll make sure the cause is:
- within the first 20% of the frozen frame
- at least 50% of the frozen frame duration
- contained inside the frozen frame with a 5% margin at the start and end
- an application frame

### DIFF
--- a/internal/occurrence/frame_drop.go
+++ b/internal/occurrence/frame_drop.go
@@ -1,6 +1,9 @@
 package occurrence
 
 import (
+	"math"
+	"time"
+
 	"github.com/getsentry/vroom/internal/frame"
 	"github.com/getsentry/vroom/internal/nodetree"
 	"github.com/getsentry/vroom/internal/profile"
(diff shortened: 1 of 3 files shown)


internal/occurrence/frame_drop.go
```diff
@@ -12,10 +15,46 @@ type (
...
+func newFrozenFrameStats(endNS uint64, durationNS float64) frozenFrameStats {
+	margin := uint64(math.Max(durationNS*marginPercent, float64(10*time.Millisecond)))
+	s := frozenFrameStats{
+		endNS:         endNS + margin,
+		durationNS:    uint64(durationNS),
+		minDurationNS: uint64(durationNS * minFrameDurationPercent),
+	}
+	if endNS >= (s.durationNS + margin) {
+		s.startNS = endNS - s.durationNS - margin
+	}
```
armcknight, line 37: Do we only allow a function frame to have started before the frozen GPU frame if the frozen frame's end is before or after (sorry can't tell which) the function frame's end? Shouldn't those be independent criteria?

armcknight (approved):
Approach is correct, just a question about the algorithm but otherwise things look good!

internal/occurrence/frame_drop.go
phacops: They are independent. We basically remove or add 5% of the frozen frame duration. Whether the end of the frame is before/after the real end of the frozen frame won't influence the check at the start.

Given a frozen from lasting 100ms, starting at 100ms in a profile, we'll consider frames potentially causing drops from 95ms to 105ms.

PR #309, opened 2023-09-08
Temp fix and switch to tunnel script

mattgauntseo-sentry:
#skip-changelog

NOTE: The `for i in $(gcloud compute os-login ssh-keys list | grep -v FINGERPRINT); do echo $i; gcloud compute os-login ssh-keys remove --key $i; done` shouldn't be needed after it's run once and cleared out any old ssh keys.


ellisonmarks (approved):
LGTM

armcknight (approved):

PR #312, opened 2023-09-13
feat: Use pipedreams to deploy

phacops:


.github/workflows/validate-pipelines.yml
```diff
@@ -1,54 +0,0 @@
...
-            - uses: actions/checkout@v3
-            - id: 'auth'
-              uses: google-github-actions/auth@v1
-              with:
-                workload_identity_provider: 'projects/868781662168/locations/global/workloadIdentityPools/prod-github/providers/github-oidc-pool'
-                service_account: 'gha-gocd-api@sac-prod-sa.iam.gserviceaccount.com'
-                token_format: 'id_token'
-                id_token_audience: '610575311308-9bsjtgqg4jm01mt058rncpopujgk3627.apps.googleusercontent.com'
-                id_token_include_email: true
-            - uses: getsentry/action-validate-gocd-pipelines@v1
```
mattgauntseo-sentry, line 50: I would recommend keeping this action since it passes the generated pipelines to the GoCD server which is the best validation we can offer at the moment.

mattgauntseo-sentry: This is my bad for not updating the docs (which I've done now) - https://www.notion.so/sentry/Pipedreams-in-GoCD-with-Jsonnet-430f46b87fa14650a80adf6708b088d9?pvs=4#517bfd0c0cd044b2a72914dcc4cb1cd9

gocd/templates/pipelines/vroom.libsonnet
```diff
@@ -0,0 +1,72 @@
+local gocdtasks = import 'github.com/getsentry/gocd-jsonnet/libs/gocd-tasks.libsonnet';
+
+function(region) {
+  environment_variables: {
+    // SENTRY_REGION is used by the dev-infra scripts to connect to GKE
+    SENTRY_REGION: region,
+    GITHUB_TOKEN: '{{SECRET:[devinfra-github][token]}}',
+    GCP_PROJECT: 'internal-sentry',
```
mattgauntseo-sentry, line 8: It might be worth removing this here to avoid any possible mix up with the environment variables set by `eval $(/devinfra/scripts/regions/project_env_vars.py --region="${SENTRY_REGION}")`

BUT assuming you want to use the image internal-sentry for all regions, we should update the cloudbuild bash to use that value.

gocd/templates/bash/check-cloudbuild.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+/devinfra/scripts/checks/googlecloud/checkcloudbuild.py \
+	${GO_REVISION_VROOM_REPO} \
+	${GCP_PROJECT} \
```
mattgauntseo-sentry, line 5: See comment regarding GCP_PROJECT on the vroom.libsonnet file.

gocd/templates/pipelines/vroom.libsonnet
```diff
@@ -0,0 +1,72 @@
...
+            tasks: [
+              gocdtasks.script(importstr '../bash/deploy.sh'),
+              gocdtasks.script(importstr '../bash/wait-canary.sh'),
+            ],
+          },
+        },
+      },
+    },
+    {
+      deploy_primary: {
```
mattgauntseo-sentry, line 55:
```suggestion
      deploy-primary: {
```

The rollback pipeline is configured for `deploy-primary` stage and this matches getsentry pipelines.

gocd/templates/vroom.jsonnet
```diff
@@ -0,0 +1,31 @@
+local vroom = import './pipelines/vroom.libsonnet';
+local pipedream = import 'github.com/getsentry/gocd-jsonnet/libs/pipedream.libsonnet';
+
+local pipedream_config = {
+  name: 'vroom',
+  auto_deploy: true,
```
mattgauntseo-sentry, line 6: Optional: We default to auto-deploy, so this can be removed if you like.

gocd/templates/pipelines/vroom.libsonnet
```diff
@@ -0,0 +1,72 @@
+local gocdtasks = import 'github.com/getsentry/gocd-jsonnet/libs/gocd-tasks.libsonnet';
+
+function(region) {
+  environment_variables: {
```
mattgauntseo-sentry, line 4: You can add `SKIP_CANARY_CHECKS: false` here so that in the GoCD UI can change it to true for a pipeline run.

mattgauntseo-sentry (changes requested):

gocd/templates/pipelines/vroom.libsonnet
phacops: I remove the variable and just hard-coded the project name. I think we have to keep this image for all regions as we're not pushing it anywhere else just yet, but maybe we'll change that as we have a EU region as we might want to use a different mirror (`eu.gcr.io`vs `us.gcr.io`)?

phacops:
@mattgauntseo-sentry Do we have a way to test this? Or maybe just merging, seeing if it works and reverting is fine with GoCD?

mattgauntseo-sentry:
The Validate GoCD Pipelines is the best test we have TBH - that action sends the generated pipeline to GoCD to see if it can accept the pipeline.

There still can be some issues, but they tend to be more nuanced, and in those cases commiting and reverting is probably the easiest.

.github/workflows/lint-deployment-pipelines.yaml
```diff
@@ -0,0 +1,18 @@
+name: Lint Deployment Pipelines
```
mattgauntseo-sentry, line 1: TBH you can remove this workflow - the same step is run in the validation-pipelines workflow on line 51

mattgauntseo-sentry (approved):

gocd/templates/jsonnetfile.json
```diff
@@ -0,0 +1,15 @@
...
+  "version": 1,
+  "dependencies": [
+    {
+      "source": {
+        "git": {
+          "remote": "https://github.com/getsentry/gocd-jsonnet.git",
+          "subdir": "libs"
+        }
+      },
+      "version": "v2.2.1"
```
mattgauntseo-sentry, line 11: If you update this to `v2.2.2` (and run `make gocd` to update jsonnetfile.lock.json) it'll have the fix for the rollback pipeline

PR #313, opened 2023-09-20
Bump go sdk

cleptric:
Bump the Go SDK to [v0.24.1](https://github.com/getsentry/sentry-go/releases/tag/v0.24.1) as well as remove the `TracesSampler` in favour of using [`InboundFilters`](https://docs.sentry.io/product/data-management-settings/filtering/#transactions-coming-from-healthcheck).

### DIFF
--- a/cmd/vroom/main.go
+++ b/cmd/vroom/main.go
@@ -173,15 +173,10 @@ func main() {
 	err = sentry.Init(sentry.ClientOptions{
 		Dsn:                   env.config.SentryDSN,
 		EnableTracing:         true,
+		TracesSampleRate:      1.0,
 		Environment:           env.config.Environment,
 		Release:               release,
 		BeforeSendTransaction: httputil.SetHTTPStatusCodeTag,
-		TracesSampler: func(ctx sentry.SamplingContext) float64 {
-			if ctx.Span.Name == "GET /health" {
-				return 0
-			}
-			return 1
-		},
 	})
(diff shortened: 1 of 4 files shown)


cmd/vroom/main.go
```diff
@@ -173,15 +173,10 @@ func main() {
...
+		TracesSampleRate:      1.0,
 		Environment:           env.config.Environment,
 		Release:               release,
 		BeforeSendTransaction: httputil.SetHTTPStatusCodeTag,
-		TracesSampler: func(ctx sentry.SamplingContext) float64 {
-			if ctx.Span.Name == "GET /health" {
-				return 0
-			}
-			return 1
-		},
```
cleptric, line 184: Inbound Data Filters are enabled for [vroom](https://sentry.sentry.io/settings/projects/vroom/filters/data-filters/) and filter out [`/health` by default](https://docs.sentry.io/product/data-management-settings/filtering/#transactions-coming-from-healthcheck).

Setting `TracesSampleRate: 1.0` will maintain the same sample rate.

cleptric: We could ofc make the argument not to add extra load to our servers. I'm fine with either direction.

Zylphrex: I feel like we tried this before and it didn't work? But we can definitely try it again.

Zylphrex (approved):

PR #316, opened 2023-09-21
fix: Close remaining open frames for Android call trees

phacops:


internal/profile/android.go
```diff
@@ -237,7 +238,13 @@ func (p Android) CallTrees() map[uint64][]*nodetree.Node {
...
 				continue
 			}
-			closeFrame(e.ThreadID, ts)
+			for i := len(stacks[e.ThreadID]) - 1; i >= 0; i-- {
+				closeFrame(e.ThreadID, ts)
+				n := stacks[e.ThreadID][i]
+				if n.Frame.MethodID == e.MethodID {
+					break
+				}
+			}
```
Zylphrex, line 247: This doesn't solve the issue of no close events for a thread rights? This would help if some exit events are missing but can this cause issues with recursive calls?

Zylphrex (approved):

internal/profile/android.go
phacops: As discussed offline, this will now close all open events still on the stack when we receive a close event related to an earlier open event. So before we had:
- open A
- open B
- close A (but it would close B)
- open C
- close C (but it would close B)
- a loop later would close C

This would make a bad flamegraph. Now we'll have:
- open A
- open B
- close B, close A
- open C
- close C
This will make a much better flamegraph.

We could still have missing open events which would still mess up the flamegraph.

internal/profile/android.go
```diff
@@ -237,18 +237,28 @@ func (p Android) CallTrees() map[uint64][]*nodetree.Node {
...
 				continue
 			}
-			closeFrame(e.ThreadID, ts)
+			for i := len(stacks[e.ThreadID]) - 1; i >= 0; i-- {
+				n := stacks[e.ThreadID][i]
+				if n.Frame.MethodID != e.MethodID {
+					continue
+				}
+				closeFrame(e.ThreadID, ts, i)
+				break
```
Zylphrex, line 246: Is this `break` from debugging?

internal/profile/android.go
```diff
@@ -237,18 +237,28 @@ func (p Android) CallTrees() map[uint64][]*nodetree.Node {
...
+			closeFrame(threadID, maxTimestampNS, i)
 		}
 	}
-
-	return trees
+	for _, trees := range treesByThreadID {
+		for _, root := range trees {
+			root.Close(maxTimestampNS)
+		}
+	}
```
Zylphrex, line 260: Is there a case that actually requires that you go over all the trees and close them again? Or is this just to be safe?

internal/profile/android.go
```diff
@@ -409,31 +423,30 @@ func (p Android) Speedscope() (speedscope.Output, error) {
...
 				}
 			}
-			if stack[i] != event.MethodID {
-				return speedscope.Output{}, fmt.Errorf(
-					"chrometrace: %w: ending event %v but stack for thread %v does not contain that record",
-					errorutil.ErrDataIntegrity,
-					event,
-					event.ThreadID,
-				)
+			if i >= 0 {
```
Zylphrex, line 438: If I understood this correctly, if `i < 0` then it couldn't find a matching enter event so the exit event should be skipped right? Can we leave a comment explaining that?

internal/profile/android.go
phacops: No, once we closed the frame that was on the stack, we can stop. We don't want to close the rest of the stack until we see the corresponding event, only the methods before since we probably didn't get an exit event for those methods and they're done.

internal/profile/android.go
phacops: It's to handle missing exit events we couldn't close. Depending on how the profile is, there could be a case where we miss exit events but an enter event for the next method was already emitted, like in the test case:
- Enter A
- Exit B
- Enter B

We're missing Enter B, Exit B and Exit A here. The loop before will look at the stack, see no remaining events and close nothing. We'd end up with some nodes with a 0 duration which need to be closed one way or another.

It might be too much though as it probably only for nodes towards the end as others "in the middle" of the profile wouldn't be impacted.

internal/profile/android.go
phacops: I added some comments and renamed a variable, let me know if it's clear now.

I also sped up the loop a bit, exiting right away when we wanted to skip the event and only cutting the stack if we didn't skip. Now, both call trees and speedscope use the same algorithm too.

PR #318, opened 2023-09-25
feat(statistical-detectors): Add new endpoint for regressed functions

Zylphrex:
This introduces a new endpoint to create occurrences from regressed functions. When the detectors pick up a regressed function, it'll post a batch of them to this endpoint. An example will be loaded for each regression and used to fill in the details of the regression occurrence.

New group type added in getsentry/sentry#56862

### DIFF
--- a/cmd/vroom/regressed.go
+++ b/cmd/vroom/regressed.go
@@ -20,13 +20,17 @@ func (env *environment) postRegressed(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	occurrences := occurrence.ProcessRegressedFunctions(
-		ctx,
-		hub,
-		env.storage,
-		regressedFunctions,
-		10,
-	)
+	occurrences := []*occurrence.Occurrence{}
+	for _, regressedFunction := range regressedFunctions {
+		occurrence, err := occurrence.ProcessRegressedFunction(ctx, env.storage, regressedFunction)
+		if err != nil {
+			hub.CaptureException(err)
+			continue
+		} else if occurrence == nil {
+			continue
+		}
+		occurrences = append(occurrences, occurrence)
+	}
 
 	s := sentry.StartSpan(ctx, "json.marshal")
 	data := struct {
--- a/cmd/vroom/regressed.go
+++ b/cmd/vroom/regressed.go
@@ -30,8 +30,8 @@ func (env *environment) postRegressed(w http.ResponseWriter, r *http.Request) {
 
 	s := sentry.StartSpan(ctx, "json.marshal")
 	data := struct {
-		occurrences int
-	}{occurrences: len(occurrences)}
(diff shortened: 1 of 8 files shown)


cmd/vroom/regressed.go
```diff
@@ -0,0 +1,77 @@
...
+	data := struct {
+		Occurrences int `json:"occurrences"`
+	}{Occurrences: len(occurrences)}
+	b, err := json.Marshal(data)
+	s.Finish()
+	if err != nil {
+		hub.CaptureException(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
```
phacops, line 45: We'd generate the response before even sending the occurrences to Kafka?

internal/frame/frame.go
```diff
@@ -181,3 +183,62 @@ func (f Frame) IsPythonApplicationFrame() bool {
 func (f Frame) IsPHPApplicationFrame() bool {
 	return !strings.Contains(f.Path, "/vendor/")
 }
+
+func (f Frame) Fingerprint() uint32 {
+	frameFunction := f.Function
+	framePackage := f.ModuleOrPackage()
```
phacops, line 189: We don't really need to create those variables.

internal/occurrence/regressed_frame.go
```diff
@@ -0,0 +1,131 @@
...
+	calltrees, exists := calltreesByTID[p.Transaction().ActiveThreadID]
+	if !exists {
+		return nil, errors.New("calltree not found")
+	}
+
+	s = sentry.StartSpan(ctx, "processing")
+	s.Description = "Searching for fingerprint"
+	var node *nodetree.Node
+	for _, calltree := range calltrees {
+		node = calltree.FindNodeByFingerprint(regressedFunction.Fingerprint)
```
phacops, line 67: Shouldn't you stop looking if `node` is not `nil` between call trees??

cmd/vroom/regressed.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, regressedFunction := range regressedFunctions {
+		occurrence, err := occurrence.ProcessRegressedFunction(ctx, env.storage, regressedFunction)
+		if err != nil {
+			hub.CaptureException(err)
+			continue
+		} else if occurrence == nil {
+			continue
+		}
+		occurrences = append(occurrences, occurrence)
+	}
```
phacops, line 33: Should we measure each call to `ProcessRegressedFunction`? I know we're measuring some function calls inside, but wouldn't we want a span around all this?

phacops (approved):

cmd/vroom/regressed.go
Zylphrex: I don't think there's a practical difference here but the intention is to avoid a situation where after writing to kafka, the response fails to mashal.

PR #325, opened 2023-09-28
fix(cocoa): Mark sentry frames as system

Zylphrex:
Sentry frames are known to be a system frame. So if they're detected, make sure they're not marked as an application frame.


phacops (approved):
Are we correctly classifying Sentry frames for other SDKs?

Zylphrex:
> Are we correctly classifying Sentry frames for other SDKs?

A quick spot check says other platforms look good.

PR #328, opened 2023-09-29
fix: Do not return an occurrence for unknown function or when the stack is filled with them

phacops:


internal/occurrence/frame_drop.go
```diff
@@ -141,7 +151,7 @@ func findFrameDropCauseFrame(
 	// Create a nodeStack of the current node
 	ns := &nodeStack{depth, n, nil}
 	// Check if current node if valid.
-	if stats.IsNodeStackValid(ns) {
+	if stats.IsNodeStackValid(ns) && ns.n.Frame.Function != "" {
```
Zylphrex, line 154: nit: could move this condition into `IsNodeStackValid`

Zylphrex (approved):

PR #337, opened 2023-10-17
fix: make android samples monotonic

viglia:
Wall-clock time is supposed to be monotonic. In a few rare cases we've noticed this was not the case.

Due to some overflow happening client-side in the embedded profiler, the sequence might be decreasing at certain points.

This is just a workaround to mitigate this issue (at rendering time) and avoid a front-end crash, should it happen.

### DIFF
--- a/internal/profile/android.go
+++ b/internal/profile/android.go
@@ -197,6 +197,13 @@ func getAdjustedTime(maxSecs, latest, current uint64) uint64 {
 	return maxSecs + (current - latest)
 }
 
+// Wall-clock time is supposed to be monotonic
+// in a few rare cases we've noticed this was not the case.
+// Due to some overflow happening client-side in the embedded
+// profile, the sequence might be decreasing at certain points.
+//
+// This is just a workaround to mitigate this issue, should it
+// happen.
 func (p *Android) FixSamplesTime() {
 	threadMaxSecs := make(map[uint64]uint64)
 	threadLatest := make(map[uint64]uint64)
@@ -224,6 +231,8 @@ func (p *Android) FixSamplesTime() {
 
 // CallTrees generates call trees for a given profile.
 func (p Android) CallTrees() map[uint64][]*nodetree.Node {
+	// in case wall-clock.secs is not monotonic, "fix" it
+	p.FixSamplesTime()
 	var activeThreadID uint64
 	for _, thread := range p.Threads {
 		if thread.Name == mainThread {
@@ -362,6 +371,8 @@ func (p *Android) NormalizeMethods(pi profileInterface) {
 }
 
 func (p Android) Speedscope() (speedscope.Output, error) {
+	// in case wall-clock.secs is not monotonic, "fix" it
+	p.FixSamplesTime()
 	frames := make([]speedscope.Frame, 0)
 	methodIDToFrameIndex := make(map[uint64][]int)
 	for _, method := range p.Methods {
--- a/internal/profile/android.go
+++ b/internal/profile/android.go
@@ -234,8 +234,6 @@ func (p *Android) FixSamplesTime() {
 
 // CallTrees generates call trees for a given profile.
 func (p Android) CallTrees() map[uint64][]*nodetree.Node {
-	// in case wall-clock.secs is not monotonic, "fix" it
-	p.FixSamplesTime()
 	var activeThreadID uint64
 	for _, thread := range p.Threads {
 		if thread.Name == mainThread {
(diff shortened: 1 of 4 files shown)


internal/profile/android.go
```diff
@@ -180,8 +180,59 @@ func (p Android) TimestampGetter() func(EventTime) uint64 {
...
+	threadLatest := make(map[uint64]uint64)
+	var regression bool
+
+	for i, event := range p.Events {
+		current := event.Time.Monotonic.Wall.Secs
+		if current < threadLatest[event.ThreadID] {
+			regression = true
+		}
+
+		if regression {
```
phacops, line 218: Are you trying to make it so, based on a condition, it'll kick in the logic for the rest of the events (since you never reset `regression`? I'd cut that part and just compare `current < threadLatest[event.ThreadID]`, no need for a `regression` variable.

internal/profile/android.go
```diff
@@ -180,8 +180,59 @@ func (p Android) TimestampGetter() func(EventTime) uint64 {
 	return buildTimestamp
 }
 
+func uint64Max(a, b uint64) uint64 {
```
phacops, line 183: We have a `max` function with go1.21, working with many types. Use it.

internal/profile/android.go
```diff
@@ -180,8 +180,59 @@ func (p Android) TimestampGetter() func(EventTime) uint64 {
...
+// Wall-clock time is supposed to be monotonic
+// in a few rare cases we've noticed this was not the case.
+// Due to some overflow happening client-side in the embedded
+// profiler, the sequence might be decreasing at certain points.
+//
+// This is just a workaround to mitigate this issue, should it
+// happen.
+func (p *Android) FixSamplesTime() {
+	threadMaxSecs := make(map[uint64]uint64)
+	threadLatest := make(map[uint64]uint64)
```
phacops, line 209: Maybe rename this so we understand it's not thread IDs we're storing but timestamps?

internal/profile/android.go
```diff
@@ -180,8 +180,59 @@ func (p Android) TimestampGetter() func(EventTime) uint64 {
...
+	var regression bool
+
+	for i, event := range p.Events {
+		current := event.Time.Monotonic.Wall.Secs
```
phacops, line 213: Add a condition to return immediately from the function if the clock is `CPU` or `Global` since it won't have any effect.

internal/profile/android.go
```diff
@@ -180,8 +180,59 @@ func (p Android) TimestampGetter() func(EventTime) uint64 {
...
+
+		threadLatest[event.ThreadID] = current
+		threadMaxSecs[event.ThreadID] = uint64Max(threadMaxSecs[event.ThreadID], current)
+	}
+}
+
 // CallTrees generates call trees for a given profile.
 func (p Android) CallTrees() map[uint64][]*nodetree.Node {
+	// in case wall-clock.secs is not monotonic, "fix" it
+	p.FixSamplesTime()
```
phacops, line 235: Put it in the `Normalize` function instead.

https://github.com/getsentry/vroom/blob/main/internal/profile/legacy.go#L236

phacops (commented):
What happens if the `Nanos` part of the clock is having this behavior? Shouldn't we take care of the whole clock?

internal/profile/android.go
viglia: right, I'm using `regression` to mark the fact that in the sequence there is at least 1 regression and from there on, we should always re-compute the timestamp, even when `current` IS NOT `< threadLatest[event.ThreadID]`.

We just shift everything.

Let's say we have: `1, 2, 9, 7, 11`

`7` is turned into 9+1 = 10

now the new starting point from now on is `10` so, event if `11 > 7`, what we want to store instead of `11` is `10+(11-7) = 14`.

viglia:
> What happens if the `Nanos` part of the clock is having this behavior? Shouldn't we take care of the whole clock?

@phacops I don't know if this can happen.

viglia:
> > What happens if the `Nanos` part of the clock is having this behavior? Shouldn't we take care of the whole clock?
> 
> @phacops I don't know if this can happen.

I've changed the logic to convert `secs` and `nanos` to a unique `duration` expressed in `ns` and act on that instead.

This will take care of everything and it's more appropriate than treating them independently.

internal/profile/android.go
viglia: yep, makes sense.

internal/profile/android.go
```diff
@@ -180,6 +180,58 @@ func (p Android) TimestampGetter() func(EventTime) uint64 {
...
+	threadMaxTimeNs := make(map[uint64]uint64)
+	threadLatestSampleTimeNs := make(map[uint64]uint64)
+	regression := false
+	regressionIndex := 0
+
+	for i, event := range p.Events {
+		current := (event.Time.Monotonic.Wall.Secs * 1e9) + event.Time.Monotonic.Wall.Nanos
+		if current < threadLatestSampleTimeNs[event.ThreadID] {
+			regressionIndex = i
+			regression = true
```
phacops, line 213: You don't have to 2 variables for what is the same thing. Define `regressionIndex := -1` and check if it's `regression > 0` to activate the condition.

phacops (approved):

PR #356, opened 2023-11-02
feat: Return a more specific error if profile exists

phacops:
This will use the code value as a fingerprint for all GCS errors.

### DIFF
--- a/cmd/vroom/main.go
+++ b/cmd/vroom/main.go
@@ -180,7 +180,14 @@ func main() {
 		Release:               release,
 		BeforeSendTransaction: httputil.SetHTTPStatusCodeTag,
 		BeforeSend: func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
-			if code := gcerrors.Code(hint.OriginalException); code != gcerrors.Unknown {
+			code := gcerrors.Code(hint.OriginalException)
+			switch code {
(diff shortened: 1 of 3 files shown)


cmd/vroom/main.go
```diff
@@ -180,7 +180,14 @@ func main() {
...
 		BeforeSend: func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
-			if code := gcerrors.Code(hint.OriginalException); code != gcerrors.Unknown {
+			code := gcerrors.Code(hint.OriginalException)
+			switch code {
+			// Ignore unknown or network errors as gcerrors returns a specific GCS error
+			// in case we have generic network errors, even if it didn't come from the gocloud
+			// library and we can't check for the specific gocloud error type as it's in
+			// an internal package.
+			case gcerrors.Canceled, gcerrors.DeadlineExceeded, gcerrors.Unknown, gcerrors.OK:
+			default:
```
Zylphrex, line 190: What happens to non gcerrors on open source?

Zylphrex (approved):
Just curious what happens on open source, otherwise LGTM

cmd/vroom/main.go
phacops: Errors not being those 2 network errors or a GCS-specific error will get the code `Unknown` and will be ignored as well.

PR #364, opened 2023-11-10
enhance: add frame-level platform field

viglia:
React-native sends mixed list of frames (Js and cocoa) with a platform field to help discern between those 2


internal/frame/frame.go
```diff
@@ -39,6 +39,7 @@ type (
 		Status          string `json:"status,omitempty"`
 		SymAddr         string `json:"sym_addr,omitempty"`
 		Symbol          string `json:"symbol,omitempty"`
+		Patform         string `json:"platform,omitempty"`
```
Zylphrex, line 42:
```suggestion
		Platform         string `json:"platform,omitempty"`
```

Zylphrex (approved):
Just a typo but otherwise LGTM

PR #375, opened 2023-12-07
feat(profiling): Support ingesting mixed android/js profiles for react-native

viglia:
This is related to the Sentry [PR](https://github.com/getsentry/sentry/pull/61357)


phacops (commented):
Don't we need a way to render the profile as part of the Android speedscope format? I guess we can leave it for another PR though, just thinking of the work ahead to fully "support" this.

Maybe tweak the changelog to be clear about only storing it?

phacops (approved):

viglia:
> Don't we need a way to render the profile as part of the Android speedscope format? I guess we can leave it for another PR though, just thinking of the work ahead to fully "support" this.
> 
> Maybe tweak the changelog to be clear about only storing it?

yes, that is correct. I split it in 2 different PRs. This is only about ingesting those profiles, not rendering.
I'll update the Changelog to make it clearer.

PR #376, opened 2023-12-08
chore(gocd): Bumping gocd-jsonnet version

IanWoodard:
Bumping to v2.7 of gocd-jsonnet which removes customer-5 as we are getting rid of it.

#skip-changelog


kneeyo1 (approved):
lgtm

Zylphrex (approved):

PR #386, opened 2024-01-04
feat: add support for speedscope rendering of Android reactnative profiles

viglia:
add support for speedscope rendering of Android ReactNative profiles with mixed android & JS profiles

### DIFF
--- a/internal/profile/legacy.go
+++ b/internal/profile/legacy.go
@@ -10,10 +10,12 @@ import (
 	"time"
 
 	"github.com/getsentry/vroom/internal/debugmeta"
+	"github.com/getsentry/vroom/internal/frame"
 	"github.com/getsentry/vroom/internal/measurements"
 	"github.com/getsentry/vroom/internal/metadata"
 	"github.com/getsentry/vroom/internal/nodetree"
 	"github.com/getsentry/vroom/internal/platform"
+	"github.com/getsentry/vroom/internal/sample"
 	"github.com/getsentry/vroom/internal/speedscope"
 	"github.com/getsentry/vroom/internal/timeutil"
 	"github.com/getsentry/vroom/internal/transaction"
@@ -22,8 +24,11 @@ import (
 const maxProfileDurationForCallTrees = 15 * time.Second
 
 var ErrProfileHasNoTrace = errors.New("profile has no trace")
+var member void
 
 type (
+	void struct{}
+
 	LegacyProfile struct {
 		RawProfile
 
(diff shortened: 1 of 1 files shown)


internal/profile/legacy.go
```diff
@@ -264,3 +292,199 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+	if ns < 1e9 {
+		return EventTime{
+			Monotonic: EventMonotonic{
+				Wall: Duration{
+					Secs:  0,
+					Nanos: ns,
+				},
+			},
+		}
+	}
```
Zylphrex, line 471: This if block seems redundant as the division/modulus handles this case already

internal/profile/legacy.go
```diff
@@ -264,3 +292,199 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+
+	mainTID, newMainTID := getMainThreadIDs(p.ThreadMetadata, usedTids)
+
+	for si, sample := range p.Samples {
+		sampleTID := sample.ThreadID
+		if sampleTID == mainTID {
+			sampleTID = newMainTID
+		}
+		tidLastTimeNs[sampleTID] = sample.ElapsedSinceStartNS
+		eventTime := getEventTimeFromElapsedNanoseconds(sample.ElapsedSinceStartNS)
```
Zylphrex, line 317: Is this timestamp on this JS sample relative to the same point in time as the android profile? If not, the JS and android threads will be misaligned.

internal/profile/legacy.go
```diff
@@ -264,3 +292,199 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+		i := 0
+		lastStack := tidLastStack[sampleTID]
+		currentStack := p.Stacks[sample.StackID]
+		for i < len(lastStack) && i < len(currentStack) {
+			if lastStack[i] != currentStack[i] {
+				break
+			}
+			i++
+		}
```
Zylphrex, line 326: Can we double-check here that this is iterating the stack in the correct order?

If I'm not mistaken, the array representing the stack has the leaf frame at index 0 and the main frame at the last index. If that is the case, this should be checking the tail of the 2 arrays match and generate an exit event for the remaining frames at the head of the array for the previous stack.

internal/profile/legacy.go
```diff
@@ -264,3 +292,199 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+		// at this point we've scanned through all the common frames at the bottom
+		// of the stack. For any frames left in the older stack we need to generate
+		// an "exit" event.
+		// This logic applies to all samples except the 1st
+		if si > 0 {
+			for j := len(lastStack) - 1; j >= i; j-- {
```
Zylphrex, line 332: If the previous comment applies, then this should iterate from `i to 0` or `i - 1 to 0` depending on if index `i` matched or not

internal/profile/legacy.go
```diff
@@ -264,3 +292,199 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+	methodSet := make(map[uint64]void)
+	threadSet := make(map[uint64]void)
+
+	mainTID, newMainTID := getMainThreadIDs(p.ThreadMetadata, usedTids)
+
+	for si, sample := range p.Samples {
+		sampleTID := sample.ThreadID
+		if sampleTID == mainTID {
+			sampleTID = newMainTID
+		}
```
Zylphrex, line 315: Can there be samples from other threads? I'm unfamiliar with how react native works exactly but JS typically only has 1 thread.

internal/profile/legacy.go
```diff
@@ -264,3 +292,199 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+			}
+			events = append(events, ev)
+		}
+		tidLastStack[sampleTID] = currentStack
+	} // end sample loop
+
+	// once we looped all the samples, for each thread
+	// we close all the events that are left open
+	for tid, lastStack := range tidLastStack {
+		// for the last exit events we use as elpased time
```
Zylphrex, line 373:
```suggestion
		// for the last exit events we use as elapsed time
```

internal/profile/legacy.go
```diff
@@ -264,3 +292,199 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+		// whatever the latest time was plus 10ms
+		closingTimeNs := tidLastTimeNs[tid] + 1e7
```
Zylphrex, line 375: For other platforms, we end the profile at the last sample and not extend the last sample for another 10ms. So for consistency, we should conform here.

Zylphrex: Yes, "losing" the last stack is exactly what I mean here as that's what we do for other platforms. We just use the last stack as the timestamp to close the frames from the previous stack.

internal/profile/legacy.go
```diff
@@ -264,3 +292,183 @@ func (p LegacyProfile) GetTransactionTags() map[string]string {
...
+				Time:     eventTime,
+			}
+			events = append(events, ev)
+		}
+		lastStack = currentStack
+	} // end sample loop
+
+	// we use the elapsed time since start of the latest sample
+	// to exit all the events of the previous one
+	closingTimeNs := p.Samples[len(p.Samples)-1].ElapsedSinceStartNS
```
Zylphrex, line 373: One thing to watch out for here is I'm not sure if you validate the sampled js profile within the react native profile against the guards in relay. For other platforms, we ensure that each thread has at least 1 sample or it will be removed from the profile. Without this guarantee, you can accidentally render an empty profile as the first and last sample will be the same and have the same timestamp.

Zylphrex (approved):

internal/profile/legacy.go
phacops: We do that already https://github.com/getsentry/relay/blob/master/relay-profiling/src/android.rs#L175.

PR #431, opened 2024-03-20
feat: add support to ingest metric_summaries for profile functions

viglia:
More specifically, we'll ingest with the tags, among other things, the profile_id of the profile from which we extracted the function metric, so that we'll be able to fetch some samples when necessary.

Env. Vars. already set with PR https://github.com/getsentry/ops/pull/9698

### DIFF
--- a/cmd/vroom/config.go
+++ b/cmd/vroom/config.go
@@ -10,6 +10,9 @@ type (
 		OccurrencesKafkaBrokers []string `env:"SENTRY_KAFKA_BROKERS_OCCURRENCES" env-default:"localhost:9092"`
 		OccurrencesKafkaTopic   string   `env:"SENTRY_KAFKA_TOPIC_OCCURRENCES"   env-default:"ingest-occurrences"`
 
+		SpansKafkaBrokers        []string `env:"SENTRY_KAFKA_BROKERS_SPANS" env-default:"localhost:9092"`
+		MetricsSummaryKafkaTopic string   `env:"SENTRY_KAFKA_TOPIC_METRICS_SUMMARY"   env-default:"snuba-metrics-summaries"`
+
 		ProfilingKafkaBrokers []string `env:"SENTRY_KAFKA_BROKERS_PROFILING" env-default:"localhost:9092"`
 		CallTreesKafkaTopic   string   `env:"SENTRY_KAFKA_TOPIC_CALL_TREES"  env-default:"profiles-call-tree"`
 		ProfilesKafkaTopic    string   `env:"SENTRY_KAKFA_TOPIC_PROFILES"    env-default:"processed-profiles"`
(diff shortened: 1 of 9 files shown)


cmd/vroom/utils.go
```diff
@@ -180,14 +181,26 @@ func extractMetricsFromFunctions(p *profile.Profile, functions []nodetree.CallTr
 			"os_version":     p.Metadata().DeviceOSVersion,
 		}
 		duration := float64(function.SelfTimesNS[0] / 1e6)
+		summary := MetricSummary{
+			Min:   duration,
+			Max:   duration,
+			Sum:   duration,
```
Zylphrex, line 187: I realize here we have a metric called `d:profiles/function.duration@millisecond` but it's actually reporting the self time.

cmd/vroom/config.go
```diff
@@ -10,6 +10,9 @@ type (
 		OccurrencesKafkaBrokers []string `env:"SENTRY_KAFKA_BROKERS_OCCURRENCES" env-default:"localhost:9092"`
 		OccurrencesKafkaTopic   string   `env:"SENTRY_KAFKA_TOPIC_OCCURRENCES"   env-default:"ingest-occurrences"`
 
+		SpansKafkaBrokers        []string `env:"SENTRY_KAFKA_BROKERS_SPANS" env-default:"localhost:9092"`
+		MetricsSummaryKafkaTopic string   `env:"SENTRY_KAFKA_TOPIC_METRICS_SUMMARY"   env-default:"snuba-metrics-summaries"`
```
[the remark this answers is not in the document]
viglia: yep, I've just added the PR to the description to avoid confusion.

Here it is: https://github.com/getsentry/ops/pull/9698

cmd/vroom/kafka.go
```diff
@@ -91,3 +119,49 @@ func buildProfileKafkaMessage(p profile.Profile) ProfileKafkaMessage {
...
 }
+
+func generateMetricSummariesKafkaMessageBatch(p *profile.Profile, metrics []sentry.Metric, metricsSummary []MetricSummary) ([]kafka.Message, error) {
+	if len(metrics) != len(metricsSummary) {
+		return nil, fmt.Errorf("len(metrics): %d - len(metrics_summary): %d", len(metrics), len(metricsSummary))
+	}
+	messages := make([]kafka.Message, 0, len(metrics))
+	for i, metric := range metrics {
+		// add profile_id to the metrics_summary tags
+		metric.GetTags()["profile_id"] = p.ID()
```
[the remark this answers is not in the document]
viglia: no `set` api here, because we're using some low-level structs that the GO sdk user won't deal with once the higher level APIs will be added to the SDK.

cmd/vroom/utils.go
viglia: not entirely sure how that differ from current `functions` dataset though?

We're currently storing `self_times` as `durations`. 

See [here](https://github.com/getsentry/snuba/blob/master/rust_snuba/src/processors/functions.rs#L38) and [here](https://github.com/getsentry/snuba/blob/master/snuba/snuba_migrations/functions/0001_functions.py#L38)

phacops: I get it, it's not because we're doing it that it's right. A bit unfortunate to catch it this late though.

phacops (approved):

PR #441, opened 2024-04-11
fix: exclude unsymbolicated frames from metrics ingestion

viglia:
This is especially important for js/node frames where, in case of missing symbolication, everything is incorrectly treated as an in_app frame.


internal/nodetree/nodetree.go
```diff
@@ -201,13 +206,13 @@ func shouldAggregateFrame(profilePlatform platform.Platform, frame frame.Frame)
 	}
 
 	// hard coded list of functions that we should not aggregate by
-	if functionDenyList, exists := functionDenyListByPlatform[profilePlatform]; exists {
+	if functionDenyList, exists := functionDenyListByPlatform[frame.Platform]; exists {
```
Zylphrex, line 209: Is this because every frame now contains the platform information?

internal/nodetree/nodetree.go
```diff
@@ -260,3 +269,13 @@ func (n *Node) FindNodeByFingerprint(target uint32) *Node {
 
 	return nil
 }
+
+func isSymbolicatedFrame(f frame.Frame) bool {
+	if f.Platform == platform.JavaScript || f.Platform == platform.Node {
+		if f.Data.JsSymbolicated != nil {
+			return *f.Data.JsSymbolicated
+		}
+		return true
```
Zylphrex, line 278: Why is the default `true`? If `JsSymbolicated` is not set, doesn't that imply the frame was not symbolicated?

internal/nodetree/nodetree.go
```diff
@@ -231,6 +236,10 @@ func shouldAggregateFrame(profilePlatform platform.Platform, frame frame.Frame)
 		}
 	}
 
+	if _, symbolcationSupported := symbolicationSupportedPlatforms[frame.Platform]; symbolcationSupported {
```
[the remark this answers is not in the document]
viglia: typo, thanks!

internal/nodetree/nodetree.go
viglia: So this one below, is an example of frame that was not sent to symbolicator because of missing `abs_path`.

In this case `symbolicated` field is not set at all. Apart from this case I'm not sure if there are other cases, but we can se as we go.

```
  {
      "data": {},
      "function": "[Native] hermesBuiltinApply",
      "in_app": true,
      "platform": "javascript"
  }
```

internal/nodetree/nodetree.go
Zylphrex: Do we need to keep a fallback for when the frame does not have a platform specified?

internal/nodetree/nodetree.go
Zylphrex: 👍 sounds good

internal/nodetree/nodetree.go
viglia: I don't think that's necessary.

`Platform` is not an optional field at the profile level (Sample [definition](https://github.com/getsentry/relay/blob/master/relay-profiling/src/sample/v1.rs#L240)) and we always set the platform at the frame level when processing the profile (see [here](https://github.com/getsentry/sentry/blob/master/src/sentry/profiles/task.py#L1126-L1133))

viglia: Or said in another way: we always set frame-level platform in Sentry processing. So if the frame platform were to be an empty string, it means that the profile platform was an empty string as well, if that's even possible.

Zylphrex: 👍 assuming there's no reason why we'd run an old profile from before this change through this function again (which I don't think there is), this should be okay 👍

Zylphrex (approved):

PR #458, opened 2024-05-06
fix: reuse one http client across different requests

viglia:
We should always reuse an http client across different requests as they're thread-safe and meant to be reused for better performance (connection pooling).

This will avoid creating a new one (and relative http transport) for each and every requests we make to send metrics.

Not doing this might lead to temporary port exhaustion and degradation of perfomances.

### DIFF
--- a/cmd/vroom/utils.go
+++ b/cmd/vroom/utils.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"math"
+	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


phacops (approved):

cmd/vroom/utils.go
```diff
@@ -213,7 +214,9 @@ func sendMetrics(ctx context.Context, p *profile.Profile, metrics []sentry.Metri
 	tr := sentry.NewHTTPSyncTransport()
 	tr.Timeout = 5 * time.Second
 	tr.Configure(sentry.ClientOptions{
-		Dsn: p.GetOptions().ProjectDSN,
+		Dsn:           p.GetOptions().ProjectDSN,
+		HTTPTransport: mClient.Transport,
```
phacops, line 218: Maybe check if you have to pass this.

PR #463, opened 2024-05-12
feat: Implement profile chunk storage

phacops:
This will store profile chunks appropriately.

https://github.com/getsentry/snuba/pull/5895

### DIFF
--- a/cmd/vroom/chunk.go
+++ b/cmd/vroom/chunk.go
@@ -1,0 +1,127 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"strconv"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/segmentio/kafka-go"
+	"gocloud.dev/gcerrors"
+
+	"github.com/getsentry/vroom/internal/chunk"
+	"github.com/getsentry/vroom/internal/storageutil"
+)
+
+func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	hub := sentry.GetHubFromContext(ctx)
+
+	s := sentry.StartSpan(ctx, "processing")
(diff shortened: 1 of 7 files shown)


internal/chunk/chunk.go
```diff
@@ -0,0 +1,66 @@
...
+		c.ID,
+	)
+}
+
+func (c *Chunk) StartEndTimestamps() (float64, float64) {
+	count := len(c.Profile.Samples)
+	if count == 0 {
+		return 0, 0
+	}
+	return c.Profile.Samples[0].Timestamp, c.Profile.Samples[count-1].Timestamp
```
Zylphrex, line 65: Here we're assuming the samples are in chronological order, which is desired but is that guaranteed?

cmd/vroom/chunk.go
```diff
@@ -0,0 +1,127 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	hub.Scope().SetContext("Profile metadata", map[string]interface{}{
+		"chunk_id":        c.ID,
+		"organization_id": strconv.FormatUint(c.OrganizationID, 10),
+		"profiler_id":     c.ProfilerID,
+		"project_id":      strconv.FormatUint(c.ProjectID, 10),
+		"size":            len(body),
```
Zylphrex, line 50: Just a thought but can we emit this as a metric to see the average payload size by platform?

Zylphrex (approved):

internal/chunk/chunk.go
phacops: https://github.com/getsentry/relay/pull/3588

cmd/vroom/chunk.go
phacops: We'll wait until @viglia is done with implementing metrics then. Maybe you can even beta test on this.

PR #464, opened 2024-05-15
fix correct in_app classification for [Native] js frames

viglia:
with this we'll classify js frame whose function starts with `[Native]` as system frames.

### DIFF
--- a/internal/frame/frame.go
+++ b/internal/frame/frame.go
@@ -161,6 +161,10 @@ func (f Frame) IsNodeApplicationFrame() bool {
 }
 
 func (f Frame) IsJavaScriptApplicationFrame() bool {
+	if strings.HasPrefix(f.Function, "[Native]") {
+		return false
+	}
+
 	if len(f.Path) == 0 {
 		return true
 	}
--- a/internal/frame/frame.go
+++ b/internal/frame/frame.go
@@ -161,7 +161,7 @@ func (f Frame) IsNodeApplicationFrame() bool {
 }
 
 func (f Frame) IsJavaScriptApplicationFrame() bool {
-	if strings.HasPrefix(f.Function, "[Native]") || strings.HasPrefix(f.Function, "[HostFunction]") || strings.HasPrefix(f.Function, "[GC") {
+	if strings.HasPrefix(f.Function, "[") {
 		return false
 	}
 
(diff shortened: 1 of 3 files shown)


internal/frame/frame.go
```diff
@@ -161,6 +161,10 @@ func (f Frame) IsNodeApplicationFrame() bool {
 }
 
 func (f Frame) IsJavaScriptApplicationFrame() bool {
+	if strings.HasPrefix(f.Function, "[Native]") {
```
krystofwoldrich, line 164: There are also `[GC` and `[HostFunction` prefixes that we can safely marked as non-in-app.

internal/frame/frame.go
```diff
@@ -161,6 +161,10 @@ func (f Frame) IsNodeApplicationFrame() bool {
 }
 
 func (f Frame) IsJavaScriptApplicationFrame() bool {
+	if strings.HasPrefix(f.Function, "[Native]") || strings.HasPrefix(f.Function, "[HostFunction]") || strings.HasPrefix(f.Function, "[GC") {
```
JonasBa, line 164: Do all function calls contained inside square brackets indicate calls to native code @krystofwoldrich @viglia? If so, we might want to just check that the string starts and ends with those and mark them as system

krystofwoldrich: The only use for it I've seen is `[Native]`, `[HostFunction]` (special case of native function) and `[GC Young Gen]`, which is `SuspendFrame`, from the profiler code that cant be JS frame.

https://github.com/facebook/hermes/blob/8a272222bdd97435e1c4f2bae9ad207cff8369af/lib/VM/Profiler/ChromeTraceSerializer.cpp#L237-L256

The `[Native]`, `[HostFunction]` are prefixes.
The `SuspendFrame` is wrapped `[SuspendFrameName]`.

Maybe we could check only for `[` since that is not allowed in JS functions names.

viglia: @JonasBa that's a good point.

I just want to make sure of one thing: this behaviour works for JS in general right? Because this check is applied to all the js frames, not just react ones.
So everything that starts with `[` will be labelled as system.

Even in browser `js` and `nodejs` an app frame couldn't start with `[`, correct?

JonasBa: I dont know if browser and nodejs follow the same idea, but I do know that brackets in function names are not allowed, so we should be safe

JonasBa (approved):

JonasBa (approved):

PR #465, opened 2024-05-15
ignore react-native js frame for metrics aggregation when not symbolicated

viglia:
In order to distinguish between normal js/nodejs frames and react-native frames, an `IsReactNative` field has been added to frame.Frame.

### DIFF
--- a/internal/frame/frame.go
+++ b/internal/frame/frame.go
@@ -44,7 +44,7 @@ type (
 		// IsReactNativeFrame is not exported as json since we only
 		// need it at runtime to distinguish browser/node js frame
 		// from ReactNative js frame.
-		IsReactNativeFrame bool `json:"-"`
+		IsReactNative bool `json:"-"`
 	}
 
 	Data struct {
--- a/internal/nodetree/nodetree.go
+++ b/internal/nodetree/nodetree.go
@@ -272,7 +272,7 @@ func (n *Node) FindNodeByFingerprint(target uint32) *Node {
 
 func isSymbolicatedFrame(f frame.Frame) bool {
 	// React-native case
-	if f.Platform == platform.JavaScript && f.IsReactNativeFrame {
+	if f.Platform == platform.JavaScript && f.IsReactNative {
 		if f.Data.JsSymbolicated != nil && *f.Data.JsSymbolicated {
 			return true
 		}
--- a/internal/sample/sample.go
+++ b/internal/sample/sample.go
@@ -228,7 +228,7 @@ func (p Profile) CallTrees() (map[uint64][]*nodetree.Node, error) {
 			for i := len(stack) - 1; i >= 0; i-- {
 				f := p.Trace.Frames[stack[i]]
 				if p.Runtime.Name == "hermes" {
-					f.IsReactNativeFrame = true
+					f.IsReactNative = true
 				}
 				f.WriteToHash(h)
 				fingerprint := h.Sum64()
--- a/internal/nodetree/nodetree_test.go
+++ b/internal/nodetree/nodetree_test.go
@@ -409,18 +409,18 @@ func TestIsSymbolicated(t *testing.T) {
 		{
 			name: "react-native-symbolicated",
 			frame: frame.Frame{
-				IsReactNativeFrame: true,
-				Platform:           "javascript",
-				Data:               frame.Data{JsSymbolicated: &testutil.True},
+				IsReactNative: true,
+				Platform:      "javascript",
+				Data:          frame.Data{JsSymbolicated: &testutil.True},
 			},
 			want: true,
 		},
 		{
 			name: "react-native-not-symbolicated",
 			frame: frame.Frame{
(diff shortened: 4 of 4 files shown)


internal/nodetree/nodetree.go
```diff
@@ -271,10 +271,16 @@ func (n *Node) FindNodeByFingerprint(target uint32) *Node {
...
-	if f.Platform == platform.JavaScript || f.Platform == platform.Node {
-		if f.Data.JsSymbolicated != nil {
-			return *f.Data.JsSymbolicated
+	// React-native case
+	if f.Platform == platform.JavaScript && f.IsReactNativeFrame {
+		if f.Data.JsSymbolicated != nil && *f.Data.JsSymbolicated {
+			return true
 		}
+		return false
+	} else if f.Platform == platform.JavaScript || f.Platform == platform.Node {
```
Zylphrex, line 280: This isnt checking if a non react native js frame or nodejs frame is symbolicated anymore. Is this intentional?

viglia: @Zylphrex yes, this is intentional.

As of now we don't want to do that for broswer js and node.js (see discussion on slack).

internal/frame/frame.go
```diff
@@ -41,6 +41,10 @@ type (
 		SymAddr         string            `json:"sym_addr,omitempty"`
 		Symbol          string            `json:"symbol,omitempty"`
 		Platform        platform.Platform `json:"platform,omitempty"`
+		// IsReactNativeFrame is not exported as json since we only
+		// need it at runtime to distinguish browser/node js frame
+		// from ReactNative js frame.
+		IsReactNativeFrame bool `json:"-"`
```
phacops, line 47: You should keep that private `isReactNative` so it's by default ignored by the JSON package and remove `Frame` from the name, a bit redundant.

internal/nodetree/nodetree.go
```diff
@@ -271,10 +271,16 @@ func (n *Node) FindNodeByFingerprint(target uint32) *Node {
 }
 
 func isSymbolicatedFrame(f frame.Frame) bool {
-	if f.Platform == platform.JavaScript || f.Platform == platform.Node {
-		if f.Data.JsSymbolicated != nil {
-			return *f.Data.JsSymbolicated
+	// React-native case
+	if f.Platform == platform.JavaScript && f.IsReactNativeFrame {
```
phacops, line 275: Wouldn't just checking for `IsReactNative` be enough here?

internal/sample/sample.go
```diff
@@ -227,6 +227,9 @@ func (p Profile) CallTrees() (map[uint64][]*nodetree.Node, error) {
 
 			for i := len(stack) - 1; i >= 0; i-- {
 				f := p.Trace.Frames[stack[i]]
+				if p.Runtime.Name == "hermes" {
+					f.IsReactNativeFrame = true
+				}
```
phacops, line 232: Shouldn't this happen when we normalize the profile instead of the CallTrees function?

phacops (commented):
Why do we add another field and modify the runtime to identify react native? Can't we change the platform field of the Frame at this point?

internal/frame/frame.go
viglia: I don't want to export this field as `json` to avoid storage in the payload in GCS, but I still want to access it from other packages (other than the `frame` package). 

If I declare it as lowercase I won't be able to access it from other packages.

internal/nodetree/nodetree.go
viglia: we also have `cocoa` frame that have `IsReactNative==true`.

This will let us distinguish from `js`, `cocoa` and `android` frames that belong to ReactNative profiles vs others.

If I don't check the `javascript` platform II'd end up applying this logic to `cocoa` frames as well.

internal/sample/sample.go
viglia: If we wanted to store this as part of the profile payload yes, but this is not something we want to store in GCS really.

viglia:
> Why do we add another field and modify the runtime to identify react native? 

@phacops the runtime actually is already set to `hermes` for react-native profiles (or at least for the react-native + cocoa).
The problem is with Android, which, if you remember, we ended up handling in a special way, where the higher level profile instead of being the react-native one is the android one and the js one is stored in the `js_profiles` field. That's why in that case the `hermes` runtime is missing. So I'm just re-setting it.

> Can't we change the platform field of the Frame at this point?

this is to avoid overriding the real frame platform. We could have `cocoa`, `android` and `javascript` frames, all coming from react-native. And we still want to apply some custom logic based on which frame we're dealing with. The `IsReactNative` will let us be even a bit more specific in case we need to, such as in the case of browser/node js frames vs react-native js frames.

internal/frame/frame.go
viglia: well, in that case I could always add a `getter`. Not sure which option it's better.

phacops (approved):

PR #470, opened 2024-06-05
feat: add endpoint to return a profile from a list of chunk IDs

viglia:
Given a list of `chunk_ids`, this will download and merge them into a single and coherent profile.

related to https://github.com/getsentry/sentry/pull/72099


internal/chunk/chunk_utils_test.go
```diff
@@ -0,0 +1,96 @@
...
+					},
+					Stacks: [][]int{
+						{0, 1},
+						{0, 1},
+						{2, 3},
+						{2, 3},
+					},
+					ThreadMetadata: map[string]map[string]string{"0x0000000102adc700": {"name": "com.apple.main-thread"}, "0x000000016d8fb180": {"name": "com.apple.network.connections"}},
+				},
+				Measurements: json.RawMessage(`{"first_metric":{"unit":"ms","values":[{"elapsed_since_start_ns":100,"value":1},{"elapsed_since_start_ns":200,"value":1.2}]}}`),
```
Zylphrex, line 80: For continuous profiling, we're using absolute timestamps for everything including measurements. So it should use this struct https://github.com/getsentry/relay/blob/master/relay-profiling/src/measurements.rs#L23

viglia: thanks for the heads-up!

Zylphrex (approved):

PR #476, opened 2024-06-21
support generating a flamegraph from a list of chunk IDs

viglia:


internal/flamegraph/flamegraph.go
```diff
@@ -25,6 +26,12 @@ type (
 	}
 
 	CallTrees map[uint64][]*nodetree.Node
+
+	ChunkMetadata struct {
```
Zylphrex, line 30: I think this needs a start and stop timestamp within the chunk since we're not always going to take the full chunk.

viglia: good point. I did something very similar for span-scoped `flamegraph` in the past. I can piggyback on that logic.

internal/flamegraph/flamegraph.go
```diff
@@ -322,3 +330,69 @@ func (f *flamegraph) getProfileIDsIndices(profileIDs map[string]struct{}) []int
...
+	// read the output of each tasks
+	for i := 0; i < len(chunksMetadata); i++ {
+		res := <-results
+		if res.Err != nil {
+			if errors.Is(res.Err, storageutil.ErrObjectNotFound) {
+				continue
+			}
+			if errors.Is(res.Err, context.DeadlineExceeded) {
+				return speedscope.Output{}, nil
+			}
```
Zylphrex, line 370: Thinking out loud here but if the result of 1 chunk is a deadline exceeded, and the rest succeeds, should we still return the flamegraph with the successful results anyways?

internal/flamegraph/flamegraph.go
```diff
@@ -322,3 +330,69 @@ func (f *flamegraph) getProfileIDsIndices(profileIDs map[string]struct{}) []int
...
+		}
+		for _, callTree := range callTrees {
+			addCallTreeToFlamegraph(&flamegraphTree, callTree, res.Chunk.ID)
+		}
+		countChunksAggregated++
+	}
+
+	sp := toSpeedscope(flamegraphTree, 4, projectID)
+	hub.Scope().SetTag("processed_chunks", strconv.Itoa(countChunksAggregated))
```
Zylphrex, line 396: Could be helpful to set the number of chunks requested as well. And this could be more useful as a metric/measurement (whatever is possible) than a tag because sending this as a string is not very useful.

Zylphrex (approved):

internal/flamegraph/flamegraph.go
viglia: That's a good question.

For the `flamechart` endpoint, where we want to return a merged sequence of _chunks_ as a unified profile, even if 1 chunk fails for whatever reason (not only timeout), we return an error as the final merged chunk wouldn't make sense with gaps in it.

For the `flamegraph`, being it an aggregation, I agree with you that this should not be a big issue and partial flamegraph is still better than no flamegraph. 
The reason I'm returning the error with an empty flamegraph here though, is that the context (`ctx`) we pass to the _goroutine_ is the one from the parent request, so if that operation failed due to a `context.DeadlineExceeded`, we've likely exceeded the parent timeout so the whole request timed-out at that point and returning a partial one won't make a difference at this point.

internal/flamegraph/flamegraph.go
viglia: Yes, I'll add the `requested_chunk` as well to the tags in the endpoint I'll add to handle the request (similarly to what we do for the other _flamegraph_ endpoint).


> And this could be more useful as a metric/measurement (whatever is possible)

Yes. That's better handled on the `Sentry` side of the request though as we have the utilities in the python SDK to easily create metrics.

PR #480, opened 2024-06-25
ref: add postChunk integration test

viglia:
#skip-changelog

### DIFF
--- a/cmd/vroom/chunk.go
+++ b/cmd/vroom/chunk.go
@@ -29,7 +29,9 @@ func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
 	body, err := io.ReadAll(r.Body)
 	s.Finish()
 	if err != nil {
-		hub.CaptureException(err)
+		if hub != nil {
+			hub.CaptureException(err)
+		}
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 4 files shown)


cmd/vroom/chunk.go
```diff
@@ -29,7 +29,9 @@ func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
 	body, err := io.ReadAll(r.Body)
 	s.Finish()
 	if err != nil {
-		hub.CaptureException(err)
+		if hub != nil {
```
phacops, line 32: I suppose you're doing this to not initialize Sentry in tests? Can we instead initialize Sentry in tests and just drop every event?

viglia: @phacops if we want to leave things as they are (without checking the `hub`) it's even easier than initializing `sentry` and then dropping the events. We can just inject a `hub` in the request context (in the test) and it'll work just fine. Like this:

```go
	ctx := req.Context()
	hub := sentry.NewHub(&sentry.Client{}, sentry.NewScope())
	ctx = context.WithValue(ctx, sentry.HubContextKey, hub)
        req = req.WithContext(ctx)
```

The reason why I've added the check is that I've noticed that `GetHubFromContext` can return a `nil`.

Now, since we're using (and wrapping our router with) the `sentryhttp` integration that should never happen as we should always get a hub in the request, but I think it's still good practice to make a nil check.

Zylphrex (approved):

PR #487, opened 2024-07-18
feat(flamegraph): Dual mode flamegraph endpoint

Zylphrex:
This introduces a new endpoint that generates a flamegraph compatible with both
transaction profiles and continuous profiles.

### DIFF
--- a/internal/chunk/readjob.go
+++ b/internal/chunk/readjob.go
@@ -15,16 +15,16 @@ type (
 		ProjectID      uint64
 		ProfilerID     string
 		ChunkID        string
-		Start          *uint64
-		End            *uint64
+		Start          uint64
+		End            uint64
 		Result         chan<- storageutil.ReadJobResult
 	}
 
 	ReadJobResult struct {
 		Err   error
 		Chunk Chunk
-		Start *uint64
-		End   *uint64
+		Start uint64
+		End   uint64
 	}
 )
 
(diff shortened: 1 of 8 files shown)


internal/chunk/readjob.go
```diff
@@ -15,12 +15,16 @@ type (
 		ProjectID      uint64
 		ProfilerID     string
 		ChunkID        string
-		Result         chan<- ReadJobResult
+		Start          *uint64
+		End            *uint64
```
phacops, line 19: Why pointers here?

Zylphrex: There was another entry point that didn't have Start/End defined. But you're right, we can use 0s as the empty values.

phacops (approved):

viglia:
@Zylphrex I'm trying to tie this back to the weekly recording I've just watched: this 3rd endpoint for the flamegraph is for the landing page use case correct? where we might have in the same project some SDKs still on the old profiling mode and some on the newer so we want to compute a flamegraph from both kind of data (profiles/chunks)?

And this is also why we don't have the `active_thread_id` info, correct? Since we're not browsing the trace viewer and we're not picking specific spans?

Zylphrex:
No, this endpoint will replace all flamegraph use cases in the future as we will need to operate in a mode where we support both profile types simultaneously and not rely on the frontend to be aware of the profile types.

Active thread id isn't there yet as I haven't passed it in yet in this first pass and will follow up on it next.

viglia:
ok, got it. Thanks 👍🏻

PR #491, opened 2024-07-24
ref: remove getProfileIDFromTransaction endpoint

viglia:
All the snuba querying is now done on the sentry side.

This removes unused endpoints and snuba utilities.

related to https://github.com/getsentry/sentry/pull/74835


Zylphrex (approved):
🔥

PR #493, opened 2024-07-26
feat: add endpoint to support live metrics aggregation

viglia:


internal/metrics/metrics.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+	}
+
+	functionsList := make([]nodetree.CallTreeFunction, 0, len(functions))
+	for _, function := range functions {
+		if function.SampleCount <= 1 {
+			// if there's only ever a single sample for this function in
+			// the profile, we skip over it to reduce the amount of data
+			continue
+		}
```
Zylphrex, line 159: Small nit from playing with the flamegraph data:

We should not be removing low volume samples until we're sure we have too many unique call trees. This makes for a better new user experience by showing all the data until we can't anymore.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,294 @@
...
+		} else if result, ok := res.(chunk.ReadJobResult); ok {
+			chunkCallTrees, err := result.Chunk.CallTrees(result.ThreadID)
+			if err != nil {
+				hub.CaptureException(err)
+				continue
+			}
+			intChunkCallTrees := make(map[uint64][]*nodetree.Node)
+			i := uint64(0)
+			for _, v := range chunkCallTrees {
+				// reald TID here doesn't really matter as it's then
```
Zylphrex, line 271:
```suggestion
				// real TID here doesn't really matter as it's then
```

internal/metrics/metrics.go
```diff
@@ -0,0 +1,294 @@
...
+			functions := CapAndFilterFunctions(ExtractFunctionsFromCallTrees(profileCallTrees), int(ma.MaxUniqueFunctions), true)
+			ma.AddFunctions(functions, result.Profile.ID())
+		} else if result, ok := res.(chunk.ReadJobResult); ok {
+			chunkCallTrees, err := result.Chunk.CallTrees(result.ThreadID)
+			if err != nil {
+				hub.CaptureException(err)
+				continue
+			}
+			intChunkCallTrees := make(map[uint64][]*nodetree.Node)
+			i := uint64(0)
```
Zylphrex, line 269:
```suggestion
			var i uint64
```

internal/metrics/metrics.go
```diff
@@ -0,0 +1,294 @@
...
+	"github.com/getsentry/vroom/internal/profile"
+	"github.com/getsentry/vroom/internal/storageutil"
+	"gocloud.dev/blob"
+)
+
+type (
+	FunctionsMetadata struct {
+		MaxVal   uint64
+		WorstID  string
+		Examples []string
```
Zylphrex, line 21: We need a more verbose type here to indicate where to find the example. If we mix `profile_id` and `chunk_id` together here, the frontend has no way of knowing which is which.

To properly support this, we'd want it to return either a `profile_id` or `profiler_id/chunk_id/thread_id` (possibly timestamps)

internal/metrics/metrics.go
viglia: @Zylphrex this makes sense, but I haven't touched this code, I've simply moved what we had here into the metrics `package`.

I'm ok with removing this check, but just to be clear: this is used both for the live aggregation and for the metrics ingestion so we might end up increasing the cardinality of the data in `clickhouse`.

If necessary I can add this "filtering" as an option to only apply for the metrics ingestion and not for the metrics live aggregation.

Zylphrex: Yep, I'm not suggesting we include the changes in this PR but rather we follow up so that we don't drop all single sample stacks for the flamegraph use case only (we can keep dropping for the functions use case)

viglia: 👍🏻

Zylphrex (approved):

PR #495, opened 2024-08-02
feat: ingest chunk metrics

viglia:


cmd/vroom/chunk.go
```diff
@@ -113,6 +115,47 @@ func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
 	}
 	s.Finish()
 
+	if c.Options.ProjectDSN != "" {
+		// nb.: here we don't have a specific thread ID, so we're going to ingest
+		// functions metrics from all the thread.
+		// That's ok as this data is not supposed to be transaction/span scoped,
+		// plus, we'll only retain application frames, so much of the system functions
+		// chaff will be dropped.
```
Zylphrex, line 123: One caveat I have noticed with this approach is that if the application spawns a background thread (for python), it'll end up being shown as the majority of the time spent. I don't have a good solution for this but this is perhaps something we'll want to solve before merging and getting a lot of bad data.

cmd/vroom/chunk.go
```diff
@@ -113,6 +115,47 @@ func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
...
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		intChunkCallTrees := make(map[uint64][]*nodetree.Node)
+		var i uint64
+		for _, v := range callTrees {
+			intChunkCallTrees[i] = v
+			i++
+		}
```
Zylphrex, line 138: Should we consider updating the callsite signature to accept string thread ids? this code is in quite a few places now. Or at least introduce a wrapper function that handles this signature instead of adding this everywhere.

cmd/vroom/chunk.go
viglia: I reflected a bit more on this and I think it's probably ok. I don't think we can really discriminate between a background thread and a legit thread of a multi-threaded app. 

Being the chunk of continuous profiling (and the metrics we show in the landing page) decoupled from transactions/spans I don't think we can make assumption about which threads we should choose for metrics ingestion. We should consider function metrics collected from all the threads, then, if the user knows better about its app, they can narrow down and get better/more-focused metrics selecting a specific transaction (assuming they're using performance too).

CHANGELOG.md
```diff
@@ -21,7 +21,11 @@
 - Pass thread id to calltree generation ([#492](https://github.com/getsentry/vroom/pull/492))
 - Dual mode metrics endpoint ([#493](https://github.com/getsentry/vroom/pull/493))
 - Add optional generation of metrics during flamegraph aggregation ([#494](https://github.com/getsentry/vroom/pull/494))
+<<<<<<< HEAD
```
Zylphrex, line 24: Looks like you have a bad merge here.

Zylphrex (approved):

CHANGELOG.md
viglia: ah, thanks!

PR #499, opened 2024-08-06
fix: handle js profile normalization for react+android profiles

viglia:
Same as https://github.com/getsentry/vroom/pull/464.

For react+android profiles though, the _js_ profile is stored as `jsonRawMessage` under a different field and we were not applying any normalization logic to that. This PR fixes it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-**Bug Fixes**:
-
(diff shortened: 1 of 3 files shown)


krystofwoldrich:
Awesome! Thank you. Looks good to me, but I leave the approval on someone who works with vroom more often.

Zylphrex (approved):

PR #501, opened 2024-08-08
feat(profiling): Annotate flamegraph with profile data

Zylphrex:
This annotates the flamegraph with metadata that allows us to link to a profile regardless if it's a transaction based profile or continuous profile.


internal/flamegraph/flamegraph.go
```diff
@@ -390,7 +440,7 @@ func GetFlamegraphFromChunks(
 			intervals := []utils.Interval{interval}
 			for _, callTree := range callTrees {
 				slicedTree := sliceCallTree(&callTree, &intervals)
-				addCallTreeToFlamegraph(&flamegraphTree, slicedTree, result.Chunk.ID)
+				addCallTreeToFlamegraph(&flamegraphTree, slicedTree, annotateWithProfileID(result.Chunk.ID))
```
viglia, line 443: I think here we might want to add an `ExampleMetadata` as well instead of annotating it with a `ProfileID/ChunkID`?

```go
	annotate := annotateWithProfileExample(
		utils.NewExampleFromProfilerChunk(
			result.Chunk.ProjectID,
			result.Chunk.ProfilerID,
			result.Chunk.ID,
			result.TransactionID,
			result.ThreadID,
			result.Start,
			result.End,
		),
	)

	intervals := []utils.Interval{interval}
	for _, callTree := range callTrees {
		slicedTree := sliceCallTree(&callTree, &intervals)
                addCallTreeToFlamegraph(&flamegraphTree, slicedTree, annotate)
        }
```

viglia (approved):
LGTM

PR #507, opened 2024-08-26
feat: Forward SDK info to Kafka

phacops:


.golangci.yml
```diff
@@ -16,16 +16,19 @@ linters:
...
-    - govet
+    #- gosimple
+    #- govet
     - ineffassign
     - misspell
     - nakedret
     - prealloc
     - revive
-    - staticcheck
+    #- staticcheck
```
viglia, line 26: any reason why we're disabling `staticcheck` (deprecated, checks already covered by other linter rules, etc.)? 

Also, if not needed, can we simply remove it instead of commenting it out?

phacops: The checks I commented in this PR were taking an insane amount of CPU/memory on my machine when running `pre-commit`.

I only commented them because I didn't fully investigate why and if it's a me problem or something else. I think I might have to push this and then investigate why it's taking such an insane amount of memory.

Zylphrex (approved):

viglia (approved):
I've tried to run `pre-commit run --all-files` on my mac with those enabled and it runs smoothly.

I think those are safe to re-enable and you might temporarily run `git commit --no-verify` on your machine until you figure out what's causing that on your mac.

phacops:
> I've tried to run `pre-commit run --all-files` on my mac with those enabled and it runs smoothly.

It might just be because you have some cached results already. I'd rather remove this and investigate why it's taking such a time on a new install.

PR #508, opened 2024-09-04
ref: optimize chunks download and fix flamegraph aggregation

viglia:
This PR aims at improving/addressing the following:

- chunks download optimization: if in the list of `ContinuousProfilesCandidate` that we receive we have multiple candidates with the same `chunkID` we should download the same chunk only once. 
- fix aggregation bug: currently, if we receive multiple candidates with the same `chunkID` and (partially or completely) overlapping intervals, we'll end-up double counting certain functions (both for flamegraph and metrics).

### DIFF
--- a/internal/flamegraph/flamegraph.go
+++ b/internal/flamegraph/flamegraph.go
@@ -94,7 +94,7 @@ func GetFlamegraphFromProfiles(
 					// the profile.timestamps to be consistent with the sample/node
 					// 'start' and 'end'
 					relativeIntervalsFromAbsoluteTimestamp(&spans, uint64(p.Timestamp().UnixNano()))
-					sortedSpans := mergeIntervals(&spans)
+					sortedSpans := utils.MergeIntervals(&spans)
 					for tid, callTree := range callTrees {
 						callTrees[tid] = sliceCallTree(&callTree, &sortedSpans)
 					}
@@ -572,8 +572,7 @@ func GetFlamegraphFromCandidates(
 
 			for tid, callTree := range result.CallTrees {
 				if intervals, ok := result.Intervals[tid]; ok {
-					sortedAndMergedIntervals := mergeIntervals(&intervals)
-					for _, interval := range sortedAndMergedIntervals {
+					for _, interval := range intervals {
 						intervalExample := utils.NewExampleFromProfilerChunk(
 							result.Chunk.ProjectID,
 							result.Chunk.ProfilerID,
--- a/internal/flamegraph/flamegraph.go
+++ b/internal/flamegraph/flamegraph.go
@@ -504,6 +504,7 @@ func GetFlamegraphFromCandidates(
 			ThreadID:       candidate.ThreadID,
 			Start:          candidate.Start,
 			End:            candidate.End,
+			Intervals:      candidate.Intervals,
 			Storage:        storage,
 			Result:         results,
 		}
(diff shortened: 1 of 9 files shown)


internal/flamegraph/flamegraph.go
```diff
@@ -568,18 +569,29 @@ func GetFlamegraphFromCandidates(
...
+							&tid,
+							interval.Start,
+							interval.End,
+						)
+						annotate := annotateWithProfileExample(intervalExample)
+						slicedTree := sliceCallTree(&callTree, &[]utils.Interval{interval})
+						addCallTreeToFlamegraph(&flamegraphTree, slicedTree, annotate)
+					} // end loop intervals for a given tid
+				} else { // !intervals for a tid
+					annotate := annotateWithProfileExample(example)
```
Zylphrex, line 590: If i understand this part correctly, this else block is to handle the nil tid case but you still need to slice the call tree here the way it used to

internal/utils/structs.go
```diff
@@ -82,3 +83,50 @@ func NewExampleFromProfilerChunk(
...
+		}
+		// if we already have a candidate with such chunkID, add the interval
+		if idx, ok := chunkToIdx[c.ChunkID]; ok {
+			// if there is already an interval for a given thread ID
+			// add the interval to that list
+			if _, ok := newCandidates[idx].Intervals[tid]; ok {
+				intervals := newCandidates[idx].Intervals[tid]
+				intervals = append(intervals, newInterval)
+				if c.Start != 0 && c.End != 0 {
+					newCandidates[idx].Intervals[tid] = intervals
```
Zylphrex, line 107: This isn't doing anything to fix the double counting problem, is it? Yes it's preventing the re-reading of the same chunk, but all the intervals from the candidates are simply appended without any change meaning if 2 intervals overlap, the overlap will be double counted.

viglia: correct, this method has the only purpose of "compacting" the same profiles/chunks and store all the intervals for each of them in a list.

The merge of the intervals happen only during *flamegraph* generation where we first sort the intervals with `mergeIntervals`.

Ps. in the last refactoring I had to do to resolve conflicts with the main branch I accidentally removed the `mergeIntervals` call. I've re-introduced it.

internal/flamegraph/flamegraph.go
viglia: Is there a case where we don't have a `thread_id` but we have an *interval* though?

In case we're fetching "transaction-scoped" profiles, then we need to slice the profile only to the time-span of the transaction from *t_i* to *t_n*. In this case though, since it's scoped to the transaction, we also know we'll have the `thread_id`.

The other scenario is where we're fetching pure profiles/chunks (not tied to transactions). In this case we don't have a specific transaction interval to slice on. The only interval we might want to slice on is the actual time-range provided by the user through the interface.
I think the most *fine-grained* for that is `1m` (unless they use the absolute time-picker and then in the URL they manually change the second value) and the chunks are much shorter (~5s). Is it necessary to slice in this case then?

internal/utils/structs.go
Zylphrex: Should the merge interval call happen here instead of inside the flamegraph call to keep all the interval manipulation in one place.

cmd/vroom/__debug_bin802580351
Zylphrex, line 1: what's this?

internal/flamegraph/flamegraph.go
Zylphrex: I dont know if it's necessary that we slice the call tree or not but assume that I don't know if it will always hold. So I'd rather us be consistent with the slicing to avoid future bugs

PR #509, opened 2024-09-04
ref: re-introduce pre-commit hooks previously removed

viglia:
This is meant to undo changes made to `.golangci.yml` in https://github.com/getsentry/vroom/pull/507/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R19-R26

I've run `pre-commit run --all-files` after running both `pre-commit clean` and `pre-commit gc` to remove any cached result to confirm everything runs smooth.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,14 +16,14 @@ linters:
     - gofmt
     - goimports
     #- gosec
-    #- gosimple
-    #- govet
+    - gosimple
+    - govet
     - ineffassign
     - misspell
     - nakedret
     - prealloc
     - revive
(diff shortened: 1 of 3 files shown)


phacops (changes requested):
I understand it works for you but these linters are doing something heavy so until we look at it carefully, we won't re-enable them.

This is the resources they took when I run them. First is CPU%, second is memory.
Screenshot 2024-09-05 at 10 00 49 AM
Screenshot 2024-09-05 at 10 00 38 AM

PR #513, opened 2024-09-04
fix(flamegraph): Dispatch candidates using goroutine

Zylphrex:
By dispatching candidates using a goroutine, we can immediately start reading the results of the readjobs instead of waiting until all candidates have been dispatched.


phacops (approved):

viglia (approved):

Zylphrex:
#512 seems to have had a significant enough impact that this doesn't look necessary anymore.

PR #520, opened 2024-10-08
ref: stop writing metrics into generic metrics for transaction-based profiling

viglia:


phacops (changes requested):
Let's hold on for now, we're using the metrics at some places already.

PR #521, opened 2024-10-17
feat: Ingest Android profile chunks

phacops:
This doesn't support:
- generating flamegraphs
- extracting metrics

### DIFF
--- a/cmd/vroom/chunk.go
+++ b/cmd/vroom/chunk.go
@@ -21,6 +21,12 @@ import (
 	"github.com/getsentry/vroom/internal/storageutil"
 )
 
+type (
+	chunkPlatform struct {
+		Platform platform.Platform `json:"platform"`
+	}
+)
+
 func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	hub := sentry.GetHubFromContext(ctx)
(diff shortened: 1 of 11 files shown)


internal/chunk/android.go
```diff
@@ -0,0 +1,113 @@
...
+	return c.ClientSDK.Name
+}
+
+func (c AndroidChunk) SDKVersion() string {
+	return c.ClientSDK.Version
+}
+
+func (c AndroidChunk) EndTimestamp() float64 {
+	return 0
+}
```
Zylphrex, line 69: This is necessary to properly filter for it on the chunks table

internal/chunk/android.go
```diff
@@ -0,0 +1,113 @@
...
+	return c.Release
+}
+
+func (c AndroidChunk) GetRetentionDays() int {
+	return c.RetentionDays
+}
+
+func (c AndroidChunk) StartTimestamp() float64 {
+	return 0
+}
```
Zylphrex, line 105: This is necessary to properly filter for it on the chunks table

internal/chunk/sample_readjob.go
viglia, line 1: Not entirely sure why we need a `readjob` specific to `sampleChunk` here.

I think we could have a `ReadJobResult` that returns a `chunk` interface instead of one for `sampleChunk` and one for `androidChunk`?

viglia, line 1: Not entirely sure why we need a `readjob` specific to `sampleChunk` here.

I think we could have a `ReadJobResult` that returns a `chunk` interface instead of one for `sampleChunk` and one for `androidChunk`?

cmd/vroom/chunk.go
```diff
@@ -115,14 +140,17 @@ func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
 	}
 	s.Finish()
 
-	if c.Options.ProjectDSN != "" {
+	options := c.GetOptions()
+	sc, ok := c.(chunk.SampleChunk)
+
+	if options.ProjectDSN != "" && c.GetPlatform() != platform.Android && ok {
```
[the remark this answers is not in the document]
phacops: Yes, that's because we don't extract metrics for Android chunks yet.

internal/chunk/sample.go
```diff
@@ -0,0 +1,263 @@
...
+	)
+}
+
+func (c SampleChunk) StartEndTimestamps() (float64, float64) {
+	count := len(c.Profile.Samples)
+	if count == 0 {
+		return 0, 0
+	}
+	return c.Profile.Samples[0].Timestamp, c.Profile.Samples[count-1].Timestamp
+}
```
Zylphrex, line 80: Any reason why we need this method still when we have a getter for each one?

Zylphrex (approved):
I may be wrong here but there's a chance that when generating flamegraph this will error when marshalling. I'm pretty sure it's handled and skipped but when testing, can we start by using an unused org/project?

internal/chunk/sample.go
phacops: No reason, I removed it.

PR #524, opened 2024-10-29
feat: support writing functions metrics we extract from chunks into the functions dataset

viglia:
This adds support for ingesting functions metrics extracted from a _chunk_  into to the _functions_ dataset.


cmd/vroom/chunk.go
```diff
@@ -157,13 +157,38 @@ func (env *environment) postChunk(w http.ResponseWriter, r *http.Request) {
...
 		functions = metrics.CapAndFilterFunctions(functions, maxUniqueFunctionsPerProfile, true)
 		s.Finish()
 
+		// This block writes into the functions dataset
+		s = sentry.StartSpan(ctx, "json.marshal")
+		s.Description = "Marshal functions Kafka message"
+		b, err := json.Marshal(buildChunkFunctionsKafkaMessage(sc, functions))
+		s.Finish()
+		if err != nil {
+			hub.CaptureException(err)
```
Zylphrex, line 172: I know we've wrapped a lot of these in `if hub != nil` is that necessary here?

cmd/vroom/kafka.go
```diff
@@ -22,6 +23,9 @@ type (
 		RetentionDays          int                         `json:"retention_days"`
 		Timestamp              int64                       `json:"timestamp"`
 		TransactionName        string                      `json:"transaction_name"`
+		StartTimestamp         float64                     `json:"start_timestamp,omitempty"`
+		EndTimestamp           float64                     `json:"end_timestamp,omitempty"`
+		ProfilingType          string                      `json:"profiling_type,omitempty"`
```
Zylphrex, line 28: Didnt we also add the thread id?

viglia: yes, correct, but that info is stored into the individual KafkaMEssage functions (`CallTreeFunction`), not at the high level. Reason being: different functions might come from different threads.

It was added [here](https://github.com/getsentry/vroom/pull/524/files#diff-a6f5aba58c13742d930ef3a7c1a047a610293f4122aad2a70a15f963f6b097b0R108)

cmd/vroom/chunk.go
viglia: yep, good point!

internal/metrics/metrics.go
```diff
@@ -127,10 +128,15 @@ func ExtractFunctionsFromCallTrees[T comparable](
...
 	functions := make(map[uint32]nodetree.CallTreeFunction, 0)
-
-	for _, callTreesForThread := range callTrees {
+	threadID := ""
+	for tid, callTreesForThread := range callTrees {
+		if t, ok := any(tid).(string); ok {
+			threadID = t
+		} else if t, ok := any(tid).(uint64); ok {
+			threadID = strconv.FormatUint(t, 10)
+		}
```
Zylphrex, line 137: The initialization of `threadID`should happen inside the loop as this risks the possibility that the thread id from the previous iteration is carried over into the next.

```suggestion
	for tid, callTreesForThread := range callTrees {
		var threadID string
		if t, ok := any(tid).(string); ok {
			threadID = t
		} else if t, ok := any(tid).(uint64); ok {
			threadID = strconv.FormatUint(t, 10)
		}
```

internal/nodetree/nodetree.go
```diff
@@ -185,6 +187,7 @@ func (n *Node) CollectFunctions(
 					SelfTimesNS:   []uint64{selfTimeNS},
 					SumSelfTimeNS: selfTimeNS,
 					SampleCount:   n.SampleCount,
+					ThreadID:      threadID,
 				}
```
Zylphrex, line 191: What happens if the same function fingerprint appears in multiple threads? This looks like it'll write all the functions into 1 row with the first thread?

Given that we just need an example, it's not unreasonable that we pick one of the threads but if that's what we're doing, ideally we bias towards the slowest instead of the first.

viglia: Yes, that is correct.

If we wanted to ingest each functions with their thread_id if the same function appears in different thread_ids, we could have a list of functions lists instead of a single list of functions.

Then after we collected everything for each thread we could merge those into a single list of functions that is then ingested into the functions dataset.

The reason I've decided to not do that is because in that case we could, potentially, end up in cases where in the materialized view, when we pick a subset of example, pick a few examples from the same chunk but just different threads, whereas I wanted to "bias" the examples set more toward different chunks if possible.

The rational behind it was: if we're already linking the user to a chunk that has that function, once there they can just check the other threads themselves and I'd rather have other examples link to different chunks.


> Given that we just need an example, it's not unreasonable that we pick one of the threads but if that's what we're doing, ideally we bias towards the slowest instead of the first.

yes, I think it's a good point. If we pick just one thread ideally better off bias the slowest one. I can make that change

Zylphrex (approved):

PR #526, opened 2024-10-30
feat(flamegraph): Keep top N samples in flamegraph

Zylphrex:
Previously, we filtered out any samples that occurred less than 4 times. This results in a poor onboarding experience where if the user sends their first profile, it may not have many samples ands does not render anything on the view. This changes the logic to keep the top N samples instead to emphasize the most common stacks while not keeping infrequent ones when there are lots of others.


viglia (approved):
I like this. It's particularly good for when we have only a few chunks.

PR #527, opened 2024-11-06
feat(statistical-detectors): Handle profile chunks in regressed endpoint

Zylphrex:
In order to support function regressions on profile chunks, we need to change from passing a simple profile id to an example metadata so we can load the chunk containing the function.

### DIFF
--- a/internal/nodetree/nodetree.go
+++ b/internal/nodetree/nodetree.go
@@ -269,21 +269,6 @@ func (n *Node) Close(timestamp uint64) {
 	}
 }
 
-func (n *Node) FindNodeByFingerprint(target uint32) *Node {
-	if n.Frame.Fingerprint() == target {
-		return n
-	}
-
-	for _, child := range n.Children {
-		node := child.FindNodeByFingerprint(target)
-		if node != nil {
-			return node
-		}
-	}
-
-	return nil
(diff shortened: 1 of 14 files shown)


internal/occurrence/regressed_frame.go
```diff
@@ -3,132 +3,114 @@ package occurrence
...
-
-	if node == nil {
-		return nil, errors.New("fingerprint not found")
-	}
-
-	return FromRegressedFunction(p.Platform(), regressedFunction, node.Frame), nil
+	return FromRegressedFunction(platform, regressedFunction, frame), nil
 }
 
-func ProcessRegressedFunctions(
```
Zylphrex, line 81: This function was unused so may as well remove it.

internal/nodetree/nodetree.go
```diff
@@ -269,21 +269,6 @@ func (n *Node) Close(timestamp uint64) {
 	}
 }
 
-func (n *Node) FindNodeByFingerprint(target uint32) *Node {
```
Zylphrex, line 272: No longer needed. Switched from traversing the call tree to find the node to directly searching for the frame on the raw profile.

phacops (approved):

PR #530, opened 2024-11-17
feat: authentication support for kafka connection

aldy505:
Support self-hosted users better by providing auth mechanism for Kafka connections. Introduces a few environment variables: 
- `SENTRY_KAFKA_SASL_MECHANISM` (valid values: `"PLAIN"`, `"SCRAM-SHA-256"`, `"SCRAM-SHA-512"`, `""`)
- `SENTRY_KAFKA_SASL_USERNAME`
- `SENTRY_KAFKA_SASL_PASSWORD`
- `SENTRY_KAFKA_SSL_CA_PATH`
- `SENTRY_KAFKA_SSL_CERT_PATH`
- `SENTRY_KAFKA_SSL_KEY_PATH`
If the mechanism is not provided, it will default to a nil `kafka.RoundTripper`, which is the default already.

### Legal Boilerplate


aldy505:
Any updates? @Zylphrex @viglia

cmd/vroom/utils.go
```diff
@@ -1,8 +1,85 @@
...
+			}
+		}
+
+		tlsConfig = &tls.Config{
+			RootCAs:      caCertificatePool,
+			Certificates: []tls.Certificate{certs},
+		}
+	}
+
+	return &kafka.Transport{
```
viglia, line 81: Don't you need to set a `Dialer` as well here?

If writer `Transport` is `nil` [DefaultTransport](https://github.com/segmentio/kafka-go/blob/main/writer.go#L195-L198) is used,  but in this case it won't as we're explicitly setting it with `createKafkaRoundTripper`.

I think here we should first create a Transport like [here](https://github.com/segmentio/kafka-go/blob/main/transport.go#L124-L131) and then set the `SASL` and `TLS`?

aldy505: Updated.

phacops (approved):

PR #539, opened 2024-12-03
feat(devservices): Adding devservices config for vroom

IanWoodard:
Adding a devservices config for vroom that will enable it to be started using the new devservices that will replace the existing `sentry devservices` used today. This change should not impact how vroom is run today and is simply meant to add new functionality in the future.

#skip-changelog

### DIFF
--- a/devservices/config.yml
+++ b/devservices/config.yml
@@ -28,6 +28,8 @@ services:
       - devservices
     extra_hosts:
       host.docker.internal: host-gateway
+    labels:
+      - orchestrator=devservices
     restart: unless-stopped
 
 volumes:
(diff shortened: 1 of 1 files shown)


devservices/config.yml
```diff
@@ -0,0 +1,39 @@
...
+      description: Sentry's profiling service, processing and deriving data about your profiles
+  modes:
+    default: [kafka, vroom]
+services:
+  vroom:
+    image: us-central1-docker.pkg.dev/sentryio/vroom/vroom:latest
+    ports:
+      - 127.0.0.1:8085:8085
+    environment:
+      SENTRY_KAFKA_BROKERS_PROFILING: kafka-kafka-1:9092
```
hubertdeng123, line 21: I think we should stick to using either container names or localhost, but not a combination of both

viglia (approved):

PR #540, opened 2024-12-03
ref: use chunk interface instead of specific implementations

viglia:
This replaces specific usage of Chunk implementations (sample, android) with the interface.
Exception made for the merge endpoint which needs to return different outputs.

With this change we'll support Android chunks too.


cmd/vroom/chunk.go
```diff
@@ -312,26 +305,80 @@ func (env *environment) postProfileFromChunkIDs(w http.ResponseWriter, r *http.R
...
+	// appropriate utility.
+	switch chunks[0].Chunk().(type) {
+	case *chunk.SampleChunk:
+		sampleChunks := make([]chunk.SampleChunk, 0, len(chunks))
+		for _, c := range chunks {
+			sc, ok := c.Chunk().(*chunk.SampleChunk)
+			if !ok {
+				w.WriteHeader(http.StatusBadRequest)
+				return
+			}
```
Zylphrex, line 329: Can we add an error message indicating that the caller mixed sample chunks and android chunks?

cmd/vroom/chunk.go
```diff
@@ -312,26 +305,80 @@ func (env *environment) postProfileFromChunkIDs(w http.ResponseWriter, r *http.R
...
-		hub.CaptureException(err)
-		w.WriteHeader(http.StatusInternalServerError)
+	case *chunk.AndroidChunk:
+		androidChunks := make([]chunk.AndroidChunk, 0, len(chunks))
+		for _, c := range chunks {
+			ac, ok := c.Chunk().(*chunk.AndroidChunk)
+			if !ok {
+				w.WriteHeader(http.StatusBadRequest)
+				return
+			}
```
Zylphrex, line 355: Also an error messaging indicating that the caller mixed sample chunks and android chunks.

Zylphrex (approved):

PR #542, opened 2024-12-06
fix(readjob): Pass readjob result as pointer

Zylphrex:
Passing the results by value copies the profile and means we are using double the memory here. Pass it by pointer and see if it helps with the OOM problem.


internal/chunk/sample_readjob.go
```diff
@@ -62,8 +62,8 @@ type (
 
 	CallTreesReadJobResult struct {
 		Err           error
-		CallTrees     map[string][]*nodetree.Node
-		Chunk         SampleChunk
+		CallTrees     *map[string][]*nodetree.Node
```
phacops, line 65: You don't need a pointer on this, maps are always passed by reference.

Also, since we already have a pointer to the value, there's no improvement we can do here.

internal/profile/readjob.go
```diff
@@ -46,8 +46,8 @@ type (
 
 	CallTreesReadJobResult struct {
 		Err       error
-		CallTrees map[uint64][]*nodetree.Node
-		Profile   Profile
+		CallTrees *map[uint64][]*nodetree.Node
```
phacops, line 49: Same as above.

phacops (approved):

PR #552, opened 2025-01-18
ci: provide arm64 image build

aldy505:
Effort for https://github.com/getsentry/self-hosted/issues/1585

The GCB pipeline that build the image and push it to GCR will still be run, but for DockerHub, we'll take the image from GHCR (GitHub Container Registry) instead. This PR will provide ARM64 version available on GHCR and DockerHub.

Also see the comment on snuba PR, we were talking about building it once and mirroring that image everywhere (build on GitHub Actions, push to GitHub Container Registry, DockerHub and GCR): https://github.com/getsentry/snuba/pull/6800#issuecomment-2609942893

### Legal Boilerplate

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,28 +48,3 @@ jobs:
         go-version: stable
         cache: false
     - run: make test
-
-  publish-to-dockerhub:
-    name: Publish Vroom to DockerHub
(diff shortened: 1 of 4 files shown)


aldy505:
I'm going to drop this in favor of GHCR image build.

PR #553, opened 2025-01-21
fix android callTree timestamps and make them consistent with the sample (v2) format

viglia:
Android and sample format share the same CallTree struct. 
While the sample format (`v2`) set the `ts` of the nodes to as absolute `ts`, the android one is setting a relative one.
This break the callTree slicing.

An optional `SdkStartTime` has been added to optionally convert the `ts` to absolute ones.

For legacy android profiles we won't set it, hence it'll keep the relative ones.

### DIFF
--- a/internal/profile/android.go
+++ b/internal/profile/android.go
@@ -161,7 +161,9 @@ type (
 		Events    []AndroidEvent  `json:"events,omitempty"`
 		Methods   []AndroidMethod `json:"methods,omitempty"`
 		StartTime uint64          `json:"start_time,omitempty"`
-		// SdkStartTime, if set, it's an absolute ts in Ns set by the sentry SDK
+		// SdkStartTime, if set (manually), it's an absolute ts in Ns
+		// whose value comes from the chunk timestamp set by the sentry SDK.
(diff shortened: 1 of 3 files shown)


internal/profile/android.go
```diff
@@ -161,7 +161,9 @@ type (
 		Events    []AndroidEvent  `json:"events,omitempty"`
 		Methods   []AndroidMethod `json:"methods,omitempty"`
 		StartTime uint64          `json:"start_time,omitempty"`
-		Threads   []AndroidThread `json:"threads,omitempty"`
+		// SdkStartTime, if set, it's an absolute ts in Ns set by the sentry SDK
+		SdkStartTime uint64          `json:"-"`
```
Zylphrex, line 165: If this is set by the SDK, setting it to `json:"-"` means it won't persist when serialized right?

viglia: @Zylphrex little misunderstanding: I set  `SdkStartTime` myself (manually) to control the callTree generation (the ts part), but the timestamp it I set it to, comes from the  chunk `timestamp` that is set by the SDK.

phacops (approved):

PR #566, opened 2025-03-06
ref: refactor speedscope struct

viglia:
Return all the measurements (V2 and legacy) under the `measurements` field (dropping _measurement_v2_) and add a `chunkID`, similarly to the `profileID` to discern between legacy profiling and continuous profiling.

#skip-changelog


JonasBa (approved):
Thank you!

PR #576, opened 2025-03-28
ci: publish vroom to GHCR and provide a tagged release on GHCR

aldy505:
Similar work with these PRs:

- https://github.com/getsentry/relay/pull/4532
- https://github.com/getsentry/symbolicator/pull/1635
- https://github.com/getsentry/snuba/pull/6997
- https://github.com/getsentry/sentry/pull/88181

While also trying to provide a solution (or at least an alternative) for this issue: https://github.com/getsentry/self-hosted/issues/3593

### Legal Boilerplate

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,6 +49,31 @@ jobs:
         cache: false
     - run: make test
 
+  publish-to-ghcr:
+    name: Publish Vroom to GHCR
+    runs-on: ubuntu-20.04
+    if: ${{ (github.ref_name == 'main') }}
+    steps:
+      - uses: actions/checkout@v4
+      - timeout-minutes: 20
+        run: until docker pull "us-central1-docker.pkg.dev/sentryio/vroom/vroom:${{ github.sha }}" 2>/dev/null; do sleep 10; done
+      - name: Push built docker image
+        shell: bash
+        run: |
+          IMAGE_URL="us-central1-docker.pkg.dev/sentryio/vroom/vroom:${{ github.sha }}"
+          docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
+          # We push 3 tags to GHCR:
+          # first, the full sha of the commit
+          docker tag "$IMAGE_URL" ghcr.io/getsentry/vroom:${GITHUB_SHA}
+          docker push ghcr.io/getsentry/vroom:${GITHUB_SHA}
+          # second, the short sha of the commit
+          SHORT_SHA=$(git rev-parse --short "$GITHUB_SHA")
+          docker tag "$IMAGE_URL" ghcr.io/getsentry/vroom:${SHORT_SHA}
+          docker push ghcr.io/getsentry/vroom:${SHORT_SHA}
+          # finally, nightly
+          docker tag "$IMAGE_URL" ghcr.io/getsentry/vroom:nightly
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -49,6 +49,31 @@ jobs:
...
     - run: make test
 
+  publish-to-ghcr:
+    name: Publish Vroom to GHCR
+    runs-on: ubuntu-20.04
+    if: ${{ (github.ref_name == 'main') }}
+    steps:
+      - uses: actions/checkout@v4
+      - timeout-minutes: 20
+        run: until docker pull "us-central1-docker.pkg.dev/sentryio/vroom/vroom:${{ github.sha }}" 2>/dev/null; do sleep 10; done
```
BYK, line 59: Let's definitely not do this and move the actual build to GitHub Actions. Otherwise you are just wasting CPU cycles and holding up workers until the Google build is done.

Alternatively, you can just another target in `.craft.yml` which does exactly what you do here.

BYK (changes requested):

BYK:
@asottile-sentry for the review, especially on required checks.

aldy505:
> ~~@asottile-sentry for the review, especially on required checks.~~
> 
> 
> 
> Ah, the failure was due to permissions. Retried from my account, let's see if it goes through.

@BYK I believe it was due to this https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images

> When you first publish a package, the default visibility is private. To change the visibility or set access permissions, see [Configuring a package's access control and visibility](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility). You can link a published package to a repository using the user interface or command line. For more information, see [Connecting a repository to a package](https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package).

aldy505:
Hello, can you guys take a look at this? @hubertdeng123 @asottile-sentry

phacops:
We need to update the image in the deploy script as well or push to both places https://github.com/getsentry/vroom/blob/main/gocd/templates/bash/deploy.sh#L8

aldy505:
> We need to update the image in the deploy script as well or push to both places https://github.com/getsentry/vroom/blob/main/gocd/templates/bash/deploy.sh#L8

I believe this PR already push to both places.

phacops:
> I believe this PR already push to both places.

Ah, you are indeed correct and Cloud Build is responsible for this.

aldy505:
@mdtro Hey, can you take a look over the build CI? Seems like another permission issue.

.github/workflows/ci.yaml
```diff
@@ -50,51 +50,99 @@ jobs:
...
+          cache-to: type=inline
+          platforms: linux/${{ matrix.platform }}
+          tags: vroom:${{ matrix.platform }}
+          outputs: type=docker,dest=/tmp/vroom-${{ matrix.platform }}.tar
+          push: false
+
+      - name: Upload Image
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: vroom-${{ matrix.platform }}
```
BYK, line 85: I feel like we are missing out on the `<sha>` part in the tag?

BYK: Ignore this, I see it below now. But we should explain this flow in code comments.

PR #590, opened 2025-05-09
feat(helm): Testing possibility of using helm for sentry services

dmajere:
This is a bit of doodling with helm for sentry services.
Using vroom since its a relatively simple service try it on.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v2
+name: vroom
+description: A Helm chart for Vroom
+type: application
+
+version: 0.1.0
+appVersion: "1.16.0"
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -1,0 +1,132 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "ServiceName" -}}
+{{- default .Chart.Name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/* 
+Namespace 
+
+Default is set to service name, its done intentionally, 
+Release.namespace should be set to service name as well.
+We want to deploy helm chart application to non-default namespace, 
+this will allow testing before we plug it into TD routes.
+
+*/}}
+{{- define "Namespace" -}}
+{{- $service := include "ServiceName" . -}}
+{{- .Release.namespace | default $service -}}
+{{- end }}
+
+{{- define "NetworkEndpointGroup" -}}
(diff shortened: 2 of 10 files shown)


chart/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: vroom
+description: A Helm chart for Vroom
+type: application
+
+version: 0.1.0
```
dmajere, line 6: not real versions

chart/templates/_helpers.tpl
nikhars, line 1: In a world where things have moved to Helm, I would imagine that we would like to make these helpers available to service teams. Can the helpers be in a separate repository and the services be in its own repository? If so, it would be nice to see an example to that end state.

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,102 @@
+{{- $service := include "ServiceName" . -}}
```
nikhars, line 1: Is it fair to say this file is what we expect service owners to own and modify most of the times?

chart/templates/hpa.yaml
```diff
@@ -0,0 +1,51 @@
+{{- $service := include "ServiceName" . -}}
```
nikhars, line 1: I would imagine that we would want to provide hpa, vpa etc. as utility functions for all services. I am still not clear how the flow of including these utility functions for each service would look like.

Ideally, i am imagining this workflow (tell me if I am wrong)
1. User writes the most minimalistic helm to use kubernetes (this means someone can run their workload on k8s).
2. There are Sentry infrastructure components which we want to mandate in all services when they want their service to run in production. Things like cogs, hpa/vpa etc.

How does one go from step 1 to step 2.

chart/values.yaml
```diff
@@ -0,0 +1,28 @@
+project: customer-1
```
nikhars, line 1: I know you mentioned this offline, but we probably want to have an opinion to how value overrides should happen. From initial glance at helm, i did not see any overriding mechanism. Does that mean if we use helm ootb, we need values file for each override?

Or do we still want to leverage the override mechanism sentry kube provides, so build a thin wrapper around helm which can do the values override.

chart/templates/_helpers.tpl
dmajere: yes. helm supports library charts.
source of inspiration here: https://github.com/bitnami/charts/tree/main/bitnami/common
figuring out permissions for already existing internal repo right now to move some of the code there.

PR #598, opened 2025-06-17
Revert "feat(docker): add default non-root user"

hubertdeng123:
Reverts getsentry/vroom#593

This PR introduces changes that break self-hosted, integrations tests fail unless using a vroom image that is prior to this PR going in

https://github.com/getsentry/self-hosted/pull/3752/commits/efd50f13b3a0b3b3ab78579fac9840bb5a4e27dd

#skip-changelog


aldy505 (approved):

BYK:
Closed as I think we can fix this downstream.

PR #599, opened 2025-06-17
fix(docker): delete default user from Dockerfile

oioki:
This causes ownership issues described in https://github.com/getsentry/self-hosted/pull/3755

_#skip-changelog_


aldy505:
This works. @BYK @hubertdeng123

PR #604, opened 2025-06-24
fix: Classify macOS frames as application frames

phacops:


noahsmartin (approved):

internal/packageutil/package.go
```diff
@@ -25,7 +25,8 @@ func IsCocoaApplicationPackage(p string) bool {
 	return strings.HasPrefix(p, "/private/var/containers") ||
 		strings.HasPrefix(p, "/var/containers") ||
 		strings.Contains(p, "/Developer/Xcode/DerivedData") ||
-		strings.Contains(p, "/data/Containers/Bundle/Application")
+		strings.Contains(p, "/data/Containers/Bundle/Application") ||
```
armcknight, line 28: should this be prefix instead of contain?
```suggestion
		strings.HasPrefix(p, "/data/Containers/Bundle/Application") ||
```

armcknight: I can't suggest the line above since it's not in the diff but same applies there

armcknight (approved):
Approved with question

internal/packageutil/package.go
phacops: I won't touch this logic since we're not 100% certain they are actually prefix (the line above is not for example).

PR #671, opened 2026-06-22
Match common repo_link format

chromy:
Normally we use the `https://` format for `repo_link` rather than `git@origin:`

So:
```
https://github.com/getsentry/sentry-shared-kafka.git
```

Rather than:
```
git@github.com:getsentry/sentry-shared-kafka.git
```

This has caused an issue in the past for example see https://github.com/getsentry/devservices/pull/322

#skip-changelog

### DIFF
--- a/devservices/config.yml
+++ b/devservices/config.yml
@@ -7,7 +7,7 @@ x-sentry-service-config:
       remote:
         repo_name: sentry-shared-kafka
         branch: main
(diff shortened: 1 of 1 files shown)


markushi (approved):
LGTM! Thanks for doing this, I stumbled upon the same issue recently in combination with devservices.

PR #672, opened 2026-07-03
feat(chunk): Add version dispatch for Android trace profiles

0xadam-brown:
Relay will soon tag Android trace profiles with version `2.android-trace` instead of omitting the version field. I've updated UnmarshalJSON so that any profiles with the new version will continue to be routed to AndroidChunk. The existing no-version path has been preserved for backward compatibility.

I've also added a Version field to AndroidChunk so the version string survives JSON round-trips through the read path (e.g. getRawChunk). Legacy chunks without a version are unaffected thanks to omitempty.

### Legal Boilerplate

#skip-changelog

### DIFF
--- a/internal/chunk/android.go
+++ b/internal/chunk/android.go
@@ -27,6 +27,7 @@ type (
 		Environment string              `json:"environment"`
 		Platform    platform.Platform   `json:"platform"`
 		Release     string              `json:"release"`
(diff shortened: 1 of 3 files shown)


markushi (approved):
LGTM!

manessaraj (approved):

